PR #56, opened 2020-06-07
Feature/bb spi and sd

XarkLabs:
Here are some Linux fixes, cstdlib additions and minor tweaks along with a rosco_m68k dhrystone port (still without optimization, so low-ball score).  Note that dhrystone still uses full libgcc.a "local path" until the -lgcc path issue is resolved (I put a note in Makefile).  I rebuilt all the software projects and verified everything builds and I didn't cause any obvious problems.

### DIFF
--- a/code/software/dhrystone/Makefile
+++ b/code/software/dhrystone/Makefile
@@ -1,0 +1,61 @@
+# Make rosco_m68k dhrystone
+
+# Copyright (c)2020 Ross Bamford
+# See LICENSE
+
+# Hacked by Xark for dhrystone
+
+EXTRA_CFLAGS?=
+SYSINCDIR?=../libs/build/include
+SYSLIBDIR?=../libs/build/lib
+# TODO: Figure out why -lgcc causes: m68k-elf-ld: cannot find -lgcc
+#       Until then, workaround is to point to libgcc.a on your system below. -Xark
+LIBS=-lprintf -lcstdlib -lmachine -lrtlsupport -lstart_serial \
+	/home/linuxbrew/.linuxbrew/Cellar/gcc-cross-m68k@7/7.5.0/lib/gcc/m68k-elf/7.5.0/m68000/libgcc.a
+DEFINES=-DROSCO_M68K
+CFLAGS=-std=c11 -ffreestanding -nostartfiles -Wall -pedantic -Werror		\
+				-I$(SYSINCDIR) -mcpu=68010 -march=68010 -mtune=68010						\
+				-mno-align-int -mno-strict-align -Wno-pedantic -Wno-maybe-uninitialized $(DEFINES)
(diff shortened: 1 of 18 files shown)


code/software/dhrystone/Makefile
```diff
@@ -0,0 +1,61 @@
...
+# Copyright (c)2020 Ross Bamford
+# See LICENSE
+
+# Hacked by Xark for dhrystone
+
+EXTRA_CFLAGS?=
+SYSINCDIR?=../libs/build/include
+SYSLIBDIR?=../libs/build/lib
+# TODO: Figure out why -lgcc causes: m68k-elf-ld: cannot find -lgcc
+#       Until then, workaround is to point to libgcc.a on your system below. -Xark
```
roscopeco, line 12: I think this (added to LDFLAGS) will fix this issue:

```
-L $(shell dirname `m68k-elf-gcc $CFLAGS -print-libgcc-file-name`)
```

XarkLabs: Cool, that makes some sense (as we are calling linker directly, not letting gcc do it).

XarkLabs: Okay, I have made that change and it seems fine.

XarkLabs: Let me know if I need to issue another PR or something (I am am Perforce guy at work).

roscopeco: Awesome, thanks :) Shouldn't need another PR, I'll rebase this onto develop and merge.

code/software/libs/src/cstdlib/include/ctype.h
```diff
@@ -18,7 +18,7 @@
 
 #include <stdint.h>
 
-typedef   uint32_t    size_t;
+typedef unsigned long size_t;
```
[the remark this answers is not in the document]
XarkLabs: Yes, I agree for that file.  I think I accidentally left that in there double checking that types were all matching - hard to know what gcc "really" uses since hidden in __INT_TYPE__ or similar magic and then complains about unsigned long vs int etc. (even though all the same to the CPU).  The main thing is to make sure that %d %ld and %lld all "do the right thing" for 32, 32, and 64 bits (without warning when gcc is set to "max griping").

code/software/libs/src/cstdlib/include/stdint.h
```diff
@@ -0,0 +1,34 @@
...
+ * Super-basic "stdint.h" just the main types for programs that need them
+ * ------------------------------------------------------------
+ */
+
+#ifndef _ROSCOM68K_STDINT_H
+#define _ROSCOM68K_STDINT_H
+
+#include <stddef.h>
+
+typedef signed char			int8_t;
```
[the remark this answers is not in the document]
XarkLabs: Yes...sorry about that.  I will switch this for this project (but I default to tabs size 4).  Also ideally there would be a clang-format that would format the code the way you like it (I am used to Allman indenting).  However, I will try to be a code chameleon going forward.

PR #88, opened 2020-08-17
Release/revision 1.1

XarkLabs:
Here is the new GPIO interrupt example and a few other misc tweaks for rosco_m68k as mentioned.  I am working in the release-1.1 branch (with my r1 board), but AFAIK all of this should "drop into" latest develop.
Also take a look at the new examples Makefiile and see if you like my tweaks (small stuff).  I could see sprucing up others to match if you do (or the parts you do).
I tried to keep things "clean and tidy" and matching everything else, but feel free to let me know if you would like me to tweak anything.


code/software/gpio-interrupt/gpio-interrupt.c
```diff
@@ -0,0 +1,135 @@
...
+{
+  MFP(GPDR) ^= 2;               // toggle red LED
+  gpio3_interrupts += 1;        // increment interrupt counter
+  print("<Interrupted!>");      // "interrupt" main printing
+  MFP(ISRB) = ~MFP_GPIP4_BIT;   // clear interrupt in-service bit
+}
+
+noreturn void kmain()
+{
+  // clear BSS segment as C standard requires (so bss variables start zero'd)
```
roscopeco, line 47: Hmm, the `start_serial` lib should do this for you. If it isn't, it's a bug in the lib.

code/software/gpio-interrupt/gpio-interrupt.c
```diff
@@ -0,0 +1,135 @@
...
+    // delay for 1 1/2 of a second
+    for (uint32_t delay = *tick_100Hz + 150; *tick_100Hz < delay;)
+      ;
+ 
+    loop_count++;
+
+  } while (!(MFP(RSR)&0x80)); // loop until a char is waiting in UART buffer
+
+  println("");
+  println("Glad to be of service!");  // invoke Genuine Rosco Personality(TM) module
```
roscopeco, line 131: :D

code/software/libs/src/machine/machine.S
```diff
@@ -99,3 +99,12 @@ mcHalt::
     stop    #$2700
     bra.s   mcHalt
 
+; Call ROM reset vector to "soft reset" the machine
```
roscopeco, line 102: I suspect this might cause problems with the 9958, which is quite particular about the order things happen in, but I think this is a good addition, and that should probably be fixed in the bootstrap code, maybe with a warm vs cold vector.

Given that, is it worth doing `move.l $4,A0` here? The current firmware should copy the ROM start address to that address at startup, and warm reboot could be 'fixed' by changing the address to a warm boot vector once enough init is done...

roscopeco (commented):
Looks great @XarkLabs , and I very much like the Makefile improvements. Left a couple of general comments (nothing major at all), thanks very much for your work on this :)

code/software/gpio-interrupt/gpio-interrupt.c
XarkLabs: I just removed that (again) and confirmed my "int gpio3_interrupts" (last bss in my test) starts at (e.g.) 0x813573e (IIRC :sweat_smile: ).  I will leave this out and assume you will apply "start_serial" OB1 fix (which I suspect, because it is very much like my first try at clearing which "didn't work" either).

code/software/libs/src/machine/machine.S
XarkLabs: Okay, I fully understand.  I have removed all "evidence" of the soft-reset (for now, but I do like your vector idea, perhaps using a "check" word somewhere to decide cold/warm boot - which you may already have).  I verified it still returns to loader nicely without it (I think it didn't at one point - or perhaps "noreturn" on other examples made me "worry" and led to my "tactical nuclear option" - but likely "my bad" or something).  I also added a tiny bit of code that on exit will disable GPIO interrupt and restore the vector (neither seemed needed - but seemed good "example hygiene").

XarkLabs:
Should be good to go! :slightly_smiling_face:

code/software/libs/src/machine/machine.S
XarkLabs: One other thought.  For my SPI loader  and similar utilities to be (more) "useful", ideally there would be a way to set an interrupt vector and have it stay set (and enabled) after running other programs and returning to loader.  Of course in the fullness of time it (or something similar) will likely be in my ROMs, but some kind of "resident program" mechanism seems useful - if only for developing this and similar utilities (either with them "hiding" in low-meory or maybe even "reserving" some high memory - dare I say "TSR style" :sweat_smile: ).   Seemed kind of related, so leaving comment here.

code/software/gpio-interrupt/gpio-interrupt.c
roscopeco: Somewhat interestingly, this didn't turn out to be off-by-one - it was a linkage issue. When I first looked I missed that we weren't talking about non-static globals - these go into `COMMON`, not `.BSS` - it's up to the link script to place `COMMON` correctly in the bss section, and my link scripts weren't doing that. So, `COMMON` was defaulting to being placed at the end of the `BSS` section, _after_ the `.bss_end` marker. Doing the less-than-or-equal to was making it clear to `bss_end`, and then carrying on into the COMMON.

TL;DR - I think I've fixed this. Please give it a go :)

code/software/libs/src/machine/machine.S
roscopeco: Cool. Yes I think we should add the concept of warm/cold reboots, and at that point look at adding mcReset back in to soft reset. I still kinda feel like we can just switch out the address in RAM at $4 once enough setup is done so that warm reboot makes sense, and have mcReset just jump to whatever is pointed to by $4. 

Completely agree on the ability to keep stuff resident - I think implementing warm reboot makes that all the easier - the warm reboot vector would skip all the initial copying of exception vectors and whatnot (and would leave alone any RAM area we decide to reserve).

roscopeco:
Many thanks for your work on this @XarkLabs , accepted and merged 👍

PR #101, opened 2020-08-23
Feature/large memory

XarkLabs:
Mostly finishes and polishes up some things from previous init changes.  Fix typos and tweak Makefile options.  Add "absolute symbols" from firmware to linker script so they can be used from C (without icky cast or making pointer).  Fixed up symbols for _RAM_SIZE (which also set _STACK_TOP by default) _STACK_SIZE default 16K which is used to make _HEAP_END (useful for malloc),  Also a bunch of others for _MFP_VECTORS (which worked great for removing asm from gpio-interrupts), _TIMER_100HZ and EFP and Easy68K stuff.  Cleaned up gpio-interrupt example to use new features.  Added mcDelaymsec10() to delay based on 100Hz timer.  Added asm sections.

### DIFF
--- a/code/software/libs/src/start_serial/init.S
+++ b/code/software/libs/src/start_serial/init.S
@@ -1,58 +1,66 @@
 ;------------------------------------------------------------
-;                                  ___ ___ _   
-;  ___ ___ ___ ___ ___       _____|  _| . | |_ 
+;                                  ___ ___ _
+;  ___ ___ ___ ___ ___       _____|  _| . | |_
 ; |  _| . |_ -|  _| . |     |     | . | . | '_|
-; |_| |___|___|___|___|_____|_|_|_|___|___|_,_| 
-;                     |_____|       firmware v1                 
+; |_| |___|___|___|___|_____|_|_|_|___|___|_,_|
+;     |_____|       firmware v1.2
 ;------------------------------------------------------------
-; Copyright (c)2020 Ross Bamford
-; See top-level LICENSE.md for licence information.
+; Copyright (c)2020 Ross Bamford See top-level LICENSE.md for licence
+; information.
 ;
-; This is the initialization code. The loader jumps into this
-; code after the "kernel" is received via serial.
+; This is the initialization code. The loader jumps into this code after the
+; "program" is received via serial.
 ;
-; The first section is position independent, but will be loaded
-; by the loader at $28000. This section copies the rest of the
-; loaded code to $1000, and then jumps to it.
+; The first section is position independent, but will be loaded by the
+; firmware loader at $00040000 (or $00028000 prior to firmware 1.2). This
+; section copies the rest of the loaded code to $1000, and then jumps to it.
 ;
-; The second section (after RELOCATED_START) is executed
-; next, based at $1000. This just calls out to __kinit to 
-; initialize .data and .bss, and then does a jump straight
-; into kmain (the user program).
+; The second section (after RELOCATED_START) is executed next, based at $1000.
+; This just calls out to __kinit to initialize .data and .bss, and then does a
+; jump straight into kmain (the user program).
 ;
-; All of this is depending on a bit of linker magic - see 
-; rosco_m68k_kernel.ld to see how that works. 
+; All of this is depending on a bit of linker magic - see rosco_m68k_kernel.ld
+; to see how that works.
 ;------------------------------------------------------------
     include "../../shared/equates.S"
 
     section .init
-; NOTE: Loaded at $28000, but copy loop has been made position-independent.
 
-; Stack and all registers can be trashed at this point,
-; the machine is your own...
+; NOTE: Loaded at $00040000 (or $00028000 in firmware prior to r1.2), but init
+; code is position-independent.
+
+; Stack and all registers can be trashed at this point, the machine is your
+; own...
 START::
-    bset.b  #1,MFP_GPDR                 ; Turn off red LED
+    bset.b  #1,MFP_GPDR           ; turn off red LED
 
-    move.l  #_STACK_TOP,A7              ; Reset stack (from linker script)
-    lea.l   RELOCATED_START(pc),A0      ; End of init PC-rel into A0 (source)
-    move.l  #_postinit,A1               ; target into A1 (destination)
+    move.l  #_STACK_TOP,A7        ; reset stack (from linker script)
 
-.COPY_LOOP:    
-    move.l  (A0)+,(A1)+                 ; Copy source to dest, with post-increment.
-    cmp.l   #_data_end,A1               ; Done?
-    bls.s   .COPY_LOOP                  ; Nope - loop again...
-  
-.COPY_DONE:
-    jmp     _postinit                   ; Jump to copied code
-
-RELOCATED_START:
+    lea.l   START(PC),A0          ; source addr, PC-rel. start (load addr)
+    lea.l   _init,A1              ; target addr, abs. start (run addr)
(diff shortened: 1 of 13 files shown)


code/software/libs/src/start_serial/init.S
```diff
@@ -11,8 +11,8 @@
 ; This is the initialization code. The loader jumps into this
 ; code after the "kernel" is received via serial.
 ;
-; The first section is linked to be run at $40000, where the
-; loader loads the code. This section copies the rest of the
+; The first section is position independent, but will be loaded
```
roscopeco, line 14: Hmm, is this still correct? It was changed to $40000 in the 1.2 firmware, and still looks to be set to that (see `code/firmware/rosco_m68k_v1.2/include/system.h`).

roscopeco: (Sorry, commented on wrong line, specifically referring to the $28000 on the next line).

code/software/libs/src/start_serial/link_scripts/rosco_m68k_kernel.ld
```diff
@@ -16,16 +16,45 @@
...
+/* firware absolute addresses (generally not useful to override)  */
+PROVIDE(_MFP_VECTORS    = 0x00000100);  /* MFP interrupt vector base    */
+PROVIDE(_TIMER_100HZ    = 0x0000040C);  /* firmware 100Hz timer counter */
+PROVIDE(_EASY68K_ECHOON = 0x00000410);  /* Easy68k 'echo on' flag       */
+PROVIDE(_EASY68K_PROMPT = 0x00000411);  /* Easy68k 'prompt on' flag     */
+PROVIDE(_EASY68K_SHOWLF = 0x00000412);  /* Easy68k 'LF display' flag    */
+
+PROVIDE(_EFP_TABLE      = 0x00000420);  /* Extension function ptr table */
+
+PROVIDE(_EFP_PRINT      = 0x00000420);  /* Print string to console      */
```
roscopeco, line 45: I like these :) 

Thinking out loud, is it worth adding a comment in the code that sets these up to remind that if the layout changes, this must also be changed? Maybe in `code/firmware/rosco_m68k_v1.2/equates.S` or `code/firmware/rosco_m68k_v1.2/bootstrap.S` (or both)?

code/software/libs/src/start_serial/link_scripts/rosco_m68k_program.ld
```diff
@@ -38,32 +67,40 @@ SECTIONS
...
   .data ALIGN(4) :
   {
     _data_start = .;
     *(.data*)
     _data_end = .;
-  } > KRAM
+  } > RAM
+
+  /* verify initialized data ends before load address 0x28000  */
+  ASSERT((_data_end) < 0x28000,
```
roscopeco, line 90: Is this correct, given that in 1.2 it is 0x40000 as per previous comment?

code/software/libs/src/start_serial/link_scripts/rosco_m68k_kernel.ld
XarkLabs: Yes, that sounds worth doing.  Not "ideal" to have to change things in sync multiple places, but not too many options here I am aware of (and for absolute symbols, linker script is "best" option - since can't directly declare them C).  I am mostly thinking these are kind of "bowels of system" (not directly C callable etc.), but a comment to remind of the "matched set" is always good.

code/software/libs/src/start_serial/link_scripts/rosco_m68k_program.ld
XarkLabs: Pretty sure it is.  I got this number from code, not a comment.  Also UART loader "failed" when my code got to 157KB and worked at 156KB, but I will verify 0x28000 was load address (it was the old "org" in init.S, so I think it had to be right).

code/software/libs/src/start_serial/init.S
XarkLabs: I believe 0x28000 is the load address if that is what you mean.  I did testing and that was the "org" on init.S before I removed it.  I will triple check. :slightly_smiling_face:

XarkLabs: Okay, checked.  Thanks for mentioning this as now I see it comes from system.h KERNEL_LOAD_ADDRESS.  Since the code is position independent, it doesn't affect run-time, but it does affect assert (which was extra credit for myself - since I hit it and had to "debug").  So my assert is correct for r1.1, but "conservative" for r1.2.  I will make it match r1.2 (or maybe I can add a revision symbol option...but minor).

code/software/libs/src/start_serial/link_scripts/rosco_m68k_program.ld
XarkLabs: As above, KERNEL_LOAD_ADDRESS will vary.  I will mostly assume r1.2 number (but just for assert - code is PIC)

code/software/libs/src/start_serial/init.S
XarkLabs: Hmm, in my defense, the "org" was never changed to 0x40000 I don't think (and this was confusing me WRT 1.2):
https://github.com/rosco-m68k/rosco_m68k/blame/a22b394f441e5f30ea937ef832312a9c1bce3e8d/code/software/libs/src/start_serial/init.S#L29

roscopeco: Haha yes, as discussed in Discord I take full responsibility :)

code/software/libs/src/start_serial/link_scripts/rosco_m68k_kernel.ld
roscopeco: I agree. I can't think of a nice way to have a single source-of-truth for this value in this case.

code/software/libs/src/start_serial/link_scripts/rosco_m68k_program.ld
XarkLabs: I added _LOAD_ADDRESS and _RUN_ADDRESS (0x1000) to linker script, however I am removing that assert since now init works around it (and any limitations are due to loader stack - which is planned to be resolved).

code/software/libs/src/start_serial/link_scripts/rosco_m68k_kernel.ld
XarkLabs: Comment added to keep in sync with firmware equates.S

code/software/libs/src/start_serial/init.S
XarkLabs: All cleaned up, I believe.  I re-wrote some of the comments to better reflect how things are now too.

XarkLabs:
I think this addresses comments, and improved init.S seems to be working nicely (I squinted carefully at map and disassembly, and tested 768K bin with initialized array that loaded and ran fine and double checked ehbasic, adventure work as normal).

XarkLabs:
Fixed stupid modulo 64K size copy logic bug (tricky 16-bit dbra counter - see [TRAPS FOR EXPERTS](http://www.easy68k.com/paulrsm/doc/trick68k.htm)), less tricky code, but it has the advantage it appears to actually works reliably (AFAICT). :sweat_smile:   Nothing like "dog fooding" with actual testing of these large memory changes with _actual_ large programs (LCD image tests).

roscopeco (approved):
Some great changes here, looks good :) 👍

PR #104, opened 2020-08-23
Add memory count, and use it to set stack top in stage 2 and start_serial

roscopeco:


roscopeco:
I agree, but I didn't do it because the rest of the addresses aren't in there (AFAIR). I think I felt it better to do one sweep and make all of the SDB addresses be in equates.S - happy to have a ticket to do that as a separate merge?

XarkLabs (approved):
Cool.  Makes total sense, and I agree a full sweep to get all the constants declared in as few places as possible is a good separate task.  I am approving this and you can make ticket for the "constant pass" (or I can learn how :slightly_smiling_face: ).

PR #107, opened 2020-09-02
Add debug_stub debug exception handler

XarkLabs:
This is my first pass at a small debug_stub exception handler (well under 1 KiB).  It has been working great for me and I verified that all exceptions work as expected (except bus error, not sure I can generate one on my board) and I believe all the information presented is accurate.  My thinking is this could be very handy in ROM (especially if a full debugger can't fit), but even linked in I am finding it extremely useful.


roscopeco (approved):
Looks good to me, I'd very much like to have this in ROM, but I think merging to libs is a good first step on that road :)

PR #109, opened 2020-09-08
Feature/merge 12 develop 20200908

roscopeco:


code/software/libs/src/start_serial/init.S
```diff
@@ -29,12 +29,15 @@
 ; NOTE: Loaded at $00040000 (or $00028000 in firmware prior to r1.2), but init
 ; code is position-independent.
 
-; Stack and all registers can be trashed at this point, the machine is your
-; own...
 
 START::                           ; position-independent load addr
     bset.b  #1,MFP_GPDR           ; turn off red LED
+
```
roscopeco, line 35: @XarkLabs this is the change I am referring to on Discord.

XarkLabs: What I see looks fine.  With dynamic memory size, linker can't (shouldn't) deal with STACK_TOP, RAM_SIZE or the program too big assert in linkerscript (STACK_SIZE becomes more a run-time thing so malloc knows when to stop vs ram top variable).  I will look this over again, but I think it is fine and will approve (and then continue with a bit of cleanup/fallout).

XarkLabs (approved):
Looks good (and mostly what I have already "lightly tested" in my own merge - other than 68681).

PR #115, opened 2020-09-24
Feature/float support

roscopeco:


roscopeco:
Not sure whether we want to make this change or not...

### Pros:

* Supports libgcc and floating point in printf

### Cons:

* Increases binary size a fair bit

XarkLabs (approved):
I merged this into develop locally and rebuilt everything and poked around and did some tests on my stuff.  Looks like a nice improvement to me (fixes 64-bit math too). I approve (and hope these changes make it to develop etc. soon too).

PR #121, opened 2020-09-27
Fix #114

### LINKED ISSUE #114: ehBASIC does not work with MC68681
The ehBASIC input routine is still using input routines that are hardcoded to the MFP.

This should be fixed so they use the firmware IO routines.

roscopeco:
Not an ideal fix, but better than nothing. 

Supports 68681 via a Makefile flag (and conditional assembly). This turned out to be preferable to trying to detect a 68681 in the code. This _could have_ been better if CHECKCHAR was in the EFP table, but even once that is, this would still need to work on existing 1.1 and 1.2 ROMS...

### DIFF
--- a/code/software/ehbasic/Makefile
+++ b/code/software/ehbasic/Makefile
@@ -4,7 +4,7 @@
 # See LICENSE
 
 DEFINES=-DROSCO_M68K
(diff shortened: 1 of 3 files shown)


XarkLabs (approved):
Looks good.

PR #122, opened 2020-09-28
Fix #120

### LINKED ISSUE #120: ehBASIC inserts random CR when output is via V9958
When ehBASIC is being used with 1.1 or 1.2 ROMs and output is going to the V9958 text console, random carriage-returns are inserted into the output. For example, in this screenshot a CR was output halfway through typing in line 10, however the subsequent LIST shows that this CR is not actually inserted into the line.

Screenshot 2020-09-27 at 19 51 20

(Hardware issues have been ruled out, the screenshot above is from MAME, though this also happens on real hardware).

I suspect this is an issue in the ehBASIC echo routine. If that doesn't turn up anything, another place to look might be the V9958 console driver in `vdpcon.S` (specifically, is there a character it shouldn't print?)

roscopeco:
WIP until community tested.

### DIFF
--- a/code/software/ehbasic/ehbasic.S
+++ b/code/software/ehbasic/ehbasic.S
@@ -83,9 +83,31 @@ nobrk		EQU	0				* null response to INPUT causes a break
 
 * output character to the console from register d0.b
 
+; While not strictly _to spec_ to reset cursor position on LF (it should be CR), ehBASIC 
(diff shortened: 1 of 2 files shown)


roscopeco:
This has been tested and can be merged.

XarkLabs (approved):
Looks good.

PR #124, opened 2020-10-03
Stop memcheck stomping on EFP

roscopeco:


roscopeco:
This stops memcheck using hardcoded address $430 (which breaks the EFP), and also removes ANSI move left (in favour of vanilla CR) and useless CRs from printf.

XarkLabs (approved):
Great.

PR #125, opened 2020-10-03
Feature/warmboot cleanup

XarkLabs:
Cleaned up ld script defines with new EFP and other SDB things synced machine.h.  Improved warmboot example.  Also removed a few deprecated static linker things now that dynamic is the default (they didn't work, so "decoy" hazard).


roscopeco (approved):
Looks good, thanks again :)

PR #126, opened 2020-10-03
Fix #106

### LINKED ISSUE #106: Memcheck memory map is outdated
The memory map produced by Memcheck in the example programs is outdated. 

image

`[Kernel]` is no longer applicable. Instead, `[Onboard][System]` should extend from 0x0 to 0xfff and the regular `[Onboard]` block should extend from 0x1000 to 0xfffff.

roscopeco:
image


XarkLabs (approved):
Okay...I'll take your word for these.

PR #129, opened 2020-10-03
Roll to development version

roscopeco:


XarkLabs (approved):
"Snapshot" is an interesting term (vs develop or custom) but it works (and is helpful).  If we wanted to get fancy, not too hard to put crc32 in (but this is fine).

PR #131, opened 2020-10-03
Fix #112

### LINKED ISSUE #112: RS232-USB PL2302 causes problems after loading
I use no FTDI cable but an RS232 to USB cable (PL2302). With this I have an issue that after loading a program via Kermit that no ready message is shown. This due an USB reset after Kermit did load the file.
As I insert a readchar call in kmain of ehbasic or adventure then I can get the start-up message of that particular program after pressing a key. This is a dirty hack. I think the delay after loading a program is to short. (So I read from the ROM sourcecode)

roscopeco:
Increasing delay to 400000 (Approx 720ms at 10MHz).

I have decided against doing this on the 1.1 and 1.2 branches, and only in the develop branch.

### DIFF
--- a/code/firmware/rosco_m68k_development/stage2/main2.c
+++ b/code/firmware/rosco_m68k_development/stage2/main2.c
@@ -72,7 +72,7 @@ noreturn void lmain() {
     }
 
     // Wait a short while for the user's terminal to come back...
-    mcBusywait(100000);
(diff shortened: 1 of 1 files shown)


XarkLabs (approved):
Sounds good to me.  I will see how this behaves for me.

PR #150, opened 2020-11-26
Fix #142, for reals this time

### LINKED ISSUE #142: Firmware seems not recognise expanded RAM correctly fix-sd-write  branch
Did you change something? I peeked in the code (firmware) and _SDB_MEM_SIZE ($414) seems the real end of memory value. The value on $0 gives the top of internal RAM. Xarks sdfat_menu uses the INITIAL_STACK value to determine the size of memory.

roscopeco:
The reason this didn't work last time was because the pointers weren't volatile, and GCC was 'optimizing' them to a clr.b. Root cause is probably .data section pressure - I had to reduce the size of the buffer used by Easy68k trap number functions (I still think it's plenty big enough though).

### DIFF
--- a/code/firmware/rosco_m68k_development/easy68k/syscalls.c
+++ b/code/firmware/rosco_m68k_development/easy68k/syscalls.c
@@ -16,7 +16,7 @@
 #include <stdbool.h>
 #include "machine.h"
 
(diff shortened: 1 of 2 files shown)


XarkLabs (approved):
Great.

PR #151, opened 2020-11-26
Fix #92

### LINKED ISSUE #92: Firmware should only optionally use GPIOs
Currently, if the timer tick is enabled, it will always toggle GPIOs (to flash the green LED). This limits the utility of the GPIO port and makes efficient low-level programming more difficult.

While one can disable the timer completely with `bclr.b  #5,MFP_IMRB`, this also disables updating of the uptime counters, and so is not optimal. There should be a variable in the SDB that one can set to disable the GPIO toggling, while still allowing the timer to continue operating.

roscopeco:
Not sure this is still as useful as it was @XarkLabs , given we now have SPI, but I've documented the fact that SPI will be unreliable without the reserved GPIO bits set...

### DIFF
--- a/code/firmware/rosco_m68k_development/InterfaceReference.md
+++ b/code/firmware/rosco_m68k_development/InterfaceReference.md
@@ -1159,7 +1159,7 @@ exception handlers and TRAP routines.
 | 0x400   | 4    | SDB Magic (0xB105D47A)                                    |
 | 0x404   | 4    | Status code (set by exception handlers)                   |
 | 0x408   | 2    | Internal counter used by timer tick handler               |
-| 0x40A   | 2    | Reserved                                                  |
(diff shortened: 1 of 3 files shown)


XarkLabs:
Hmm, silly conflict issue:
<<<<<<< feature/fix-92
SZ_BANNER4      dc.b    "                    |_____|", $1B, "[1;30m Snapshot Firmware", $1B, "[0m", $D, $A, 0
=======
SZ_BANNER4      dc.b    "                    |_____|", $1B, "[1;37m Snapshot Firmware", $1B, "[0m", $D, $A, 0
>>>>>>> develop
Not sure the correct choice... 😅

roscopeco:
Rebased, fixed conflict :) (my bad, decided to change the color of "Snapshot Firmware" :D )

XarkLabs (approved):

PR #156, opened 2020-11-29
Add small delays to improve Kermit transfer reliability

mattuna15:
- add small delay before kermit init
- increase delay during receive
- add some error messages


roscopeco (approved):
Looks good, thanks again :)

PR #170, opened 2020-12-12
Fix #169

### LINKED ISSUE #169: Firmware 1.2 does not work on MC68000
The 1.2 (and up) firmware does not work on MC68000. This is due to the temporary bus error handler, which assumes a minimum of 68010 (and treats the stack frame accordingly).

For 68000, the easiest thing to do is just unstack the frame and jump directly (as it does not support software re-run).

As part of this fix, it's also worth bringing in the 68020 support from the development firmware.

roscopeco:


XarkLabs (approved):
Can't easily test, but looks believable. :)

PR #179, opened 2021-03-04
Fix a compilation issue with Kermit with -09

roscopeco:
# What

Correctly size a buffer in Kermit (add room for a null terminator).

# Why?

The (obsolete but still supported) v1 firmware is compiled with `-O9` instead of `-Os`. When the Kermit code is compiled with that optimisation, it exposes this error (presumably due to the specifics of the way memory is laid out, I haven't had time to dig deeply). This causes a warning on GCC 10, which causes the build to fail as we use `-Werror`.

# Do we need to do this in the 1.1 and 1.2 (and dev) Kermit?

I don't think so. At least, until we've observed incorrect behaviour, we probably shouldn't.


XarkLabs (approved):
:+1:

Issue #165, opened 2020-12-06
Trap #15 handler should not trash registers if not used as an argument


roscopeco:
Currently, the trap handlers are allowed to trash their arguments (and this is documented). However, some of the handlers also trash registers that **are not** used as arguments, for example `A0` is trashed by TRAP 15 function 14. This is annoying (to say the least) and should not happen.

* Go through all the handlers, and ensure they do not trash non-argument registers.
* Document this.

See also: #164

roscopeco:
Fix in flight for this: https://github.com/rosco-m68k/rosco_m68k/pull/180

roscopeco:
Fixed by https://github.com/rosco-m68k/rosco_m68k/pull/180

PR #180, opened 2021-03-04
Fix #164 and #165

### LINKED ISSUE #164: Trap #15 handler should not trash D0
Currently, the trap handlers are allowed to trash their arguments. Sometimes this works well (e.g. string pointing to next in an array) but often it is most annoying (e.g. trashing the function code, meaning it must be set again when e.g. printing multiple buffers consecutively).

* Go through all the handlers, and save at least the function code argument from modification.
* Document this.

roscopeco:
# What?

Stop TRAP 14 and TRAP 15 stomping on so many registers (particularly the function code register, which is super annoying when calling the same trap in a loop).

# Issues

* #164 
* #165 

# What else?

There's a bunch of other stuff in here to reduce size, as fixing the issues with the TRAPs blew the small ROM size limit. There are _technically_ breaking changes, but nothing that I think anyone will care about (e.g. the different unmapped exception handlers for different exception ranges are gone, and replaced by a single generic one. This means that the same code will be written to the system status regardless of which type of unmapped exception was triggered. I seriously doubt anyone cares about that...).

Docs updated to reflect the changes in the TRAPs.

(All tested locally with regular programs and the Easy68k demo program, plus verified small and big ROM builds still work, and unrelated stuff like SD is still good).

### DIFF
--- a/code/firmware/rosco_m68k_v1.3/InterfaceReference.md
+++ b/code/firmware/rosco_m68k_v1.3/InterfaceReference.md
@@ -443,8 +443,10 @@ console.
 Register D1.L is expected to contain the function code. Other arguments
 depend on the specific function, and are documented below.
 
-In all cases, registers used as aguments (including D1.L) are **not** 
-guaranteed to be preserved. All other registers are preserved.
+In all cases, registers used as aguments (**excluding** D1.L, which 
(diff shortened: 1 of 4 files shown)


XarkLabs (approved):
Good find & fix. :slightly_smiling_face:

PR #183, opened 2021-03-06
Replace flashy exception handlers with Xark's debug_stub

roscopeco:
# What?

For BIGROM builds (only, due to space constraints) this change replaces the old flashy exception handlers with the new ones from Xark's debug_stub library. This means that, on exception, one gets a nice dump of state and registers followed by a warm reboot rather than the old cryptic one/two/three flashes lockup.

These are installed pretty late in the boot process as they rely on output being set up and linked in the EFP table. Exceptions prior to that will keep the old flashy behaviour (as will non-BIGROMs in all cases).


XarkLabs (approved):
Tested and seems to work nicely (merged into develop and 1.3 BIGROM firmware).  I think this will be appreciated (and make debugging things easier by default). 🙂

PR #184, opened 2021-03-13
Support r2 DUART board alongside r1

roscopeco:


XarkLabs (commented):
Worked nicely for me with r1 DUART board.

XarkLabs (commented):
Hmm, still looks good (I think I clicked wrong review button...). 😅

XarkLabs (approved):

PR #188, opened 2021-03-22
Add C exceptions header, and example of C exception handler

roscopeco:
This came out of something I was playing with recently, and I think it might be generally useful (or at least a good example). 

* Adds `<machine/exception.h>` which has C struct version of the CPU's exception stack frame (For 68000/010/020 at least)
* Adds an example of using the new header (plus an assembler stub) to implement a C exception handler


XarkLabs (approved):
Excellent. 🙂

PR #190, opened 2021-03-31
Board 1.23 (QA complete) - Prep for new release

### LINKED ISSUE #148: Glue logic should use A18 or A19 for IACK, not A17
The glue logic (IC5) currently defines IACK as:

```
/IACK     = A17 * FC0 * FC1 * FC2
```

This is not correct for CPUs MC68020 and up - for those CPUs, A17 is high to indicate a Coprocessor CPU Cycle. 

For upward compatibility, this should be changed to use either A18 or A19.

(TODO: Check what later CPUs output on these pins for IACK cycles - it may be that we need multiple pins in the A16-A19 range, and if so, probably a bigger GAL...)

### LINKED ISSUE #187: Memcheck needs updating for r1.23 boards
The (forthcoming) revision 1.23 board will change (fix) the way that DTACK is handled in expansion space. 

While this is a Good Thing™ , it does incur a behavioural change in that `memcheck` will now encounter bus errors when it hits the memory hole at 0x00F00000. We need to either:

* Stop it looking at that memory, much as we do for IO space, or
* Give it a bus error handler, like the firmware uses.

I am ambivalent at this point as to which one we do. That said, and given there is talk of potentially using the memory hole for other purposes, it's _probably_ best if we avoid it altogether, though of course if we _don't_ ever do that then the bus error handler would be more future-proof.

So, in short, 🤷‍♂️

roscopeco:
This branch brings in the recent hardware changes which are now tested. I am not merging these to the 1.2 branch as I feel we should have a new release branch for FW1.3 (and a new tag for the board revision) and these changes should wind up on there.

Also:
* fix #148 
* fix #187


XarkLabs (approved):
Great. 🙂

PR #192, opened 2021-04-04
Fix #191

### LINKED ISSUE #191: BSS is bloating binary size (again)
I _suspect_ we have reintroduced the bug in which BSS bloats program binary size again somewhere along the line. However, to rule out some issue with my local environment it would be good to get verification (e.g. from @XarkLabs ) that this is reproducible on !my_machine.

If this is actually an issue, the fix is pretty trivial (make BSS `NOLOAD` in the link scripts).

roscopeco:
I'm not sure we actually need this. We should await verification of #191 before merge.


XarkLabs (approved):
Yes...

PR #196, opened 2021-04-06
Fix incorrect DUART r2 base address

roscopeco:


XarkLabs (approved):
Assumed good. :slightly_smiling_face:

PR #208, opened 2021-05-02
Add C callable mcCheckchar and checkchar define

XarkLabs:
Here is a simple C wrapper for CHECKCHAR trap, mcCheckchar() as discussed.  Also checkchar define (similar to readchar).  Also tweaked a few comments.  Works in my basic testing.


roscopeco (approved):
Looks good, thanks @XarkLabs :)

PR #209, opened 2021-05-12
Feature/start fw 2.0 development

roscopeco:
This PR brings in:

* Start development cycle for FW2.0
* Initial IDE support in stage 1 (firmware TRAPs)
* Initial IDE support in stage 2 (boot from IDE)
* CPU Info in SDB, and model/speed squawked at boot

Although there's a lot of change in here, much of it is pulling the existing code into a new development directory.


XarkLabs (approved):
Awesome. 😃

PR #210, opened 2021-05-12
First cut Xosera console

roscopeco:
This PR adds the initial cut of the Xosera text console in firmware.

* Console is quite "rough and ready" at the moment, but in testing it works well
* No advantage is taken of advanced Xosera features yet, but we can iterate on this.
* Colour is not currently supported (without direct VRAM access).
* Xosera console takes precedent over V9958 (in the unlikely event that both are present).

Proof of life:

image


XarkLabs (approved):
Nice I want to try this soon. :slightly_smiling_face:

PR #212, opened 2021-05-15
Fix MAME firmware builds

roscopeco:
Just fixing some minor build issues when building firmware for MAME.

No real code changes (except when MAME builds) so not bumping any versions.


XarkLabs (approved):
👍

PR #217, opened 2021-05-24
Feature/add mame make

goesbysteve:
Adds a new Makefile option `mame` which fills the binary to 0x10000 with 0xFF to produce a 64k rom. This means that the hex rom range in mame `./src/sbc/drivers/rosco.cpp` does not have to be altered each time the rom size changes.

I to chose a switch to more meaningful file name `rosco_m68k_mame.rom.bin` rather than the existing `serial_receive.rom.bin`.

This requires the mame branch from my fork. https://github.com/goesbysteve/mame/tree/feature/rosco_filled_64k_rom


roscopeco (approved):
Changes look good, thanks @goesbysteve :)

PR #222, opened 2021-06-04
Add (temporary) forced mode switch to 848x480 for Xosera console

roscopeco:
Temporarily adding a mode switch at init to force 848x480 mode on Xosera console.

This is intended to work with Xark's new `xosera_prog` make target in the Xosera repo, which builds both 640 and 848 modes in and defaults to 640. The console currently isn't written to detect the actual width and always assumes 848, so this change makes that mode always be selected at startup (in the fullness of time, we'll probably respond to the width reported by Xosersa, but that's a bigger change).

Tested with current Xosera in two configurations (with both 640 and 848 modes baked in, and with just 848 mode baked in).

### DIFF
--- a/code/firmware/rosco_m68k_development/main1.c
+++ b/code/firmware/rosco_m68k_development/main1.c
@@ -155,8 +155,7 @@ noreturn void main1() {
     INSTALL_EASY68K_TRAP_HANDLERS();
 
 #ifdef XOSERA_CON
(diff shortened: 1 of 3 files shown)


0xTJ (commented):
I'll leave approving to @XarkLabs, since I don't know the details of Xosera and can't test, but what's here makes sense to me.

XarkLabs (approved):
👍

PR #223, opened 2021-06-22
Don't enable video interrupts at end of bootstrap

roscopeco:
Instead of setting interrupt mask to `0` at the end of the bootstrap, this change sets it to `2`, meaning interrupts 1 and 2 (reserved for video cards) remain disabled. Later code sets it back to two anyway, this change just avoids the issue where it is 0 for a brief time, causing starvation if there is a video card installed that requires software interrupt acknowledgement to drop its IRQ line.


XarkLabs (approved):
Cool.

PR #227, opened 2021-06-26
Wait for Xosera before reconfig

XarkLabs:
This small change does a sync before reconfiguring Xosera.  I believe it is needed because typically will still be reconfiguring from hardware reset by the time the firmware init runs, so the reconfigure command was ignored (but the sync waited until initial reset finished).  Tested to properly select config #1 if present and with a "non-Xosera" config and it skips Xosera init nicely with no perceptible delay.  NOTE: After a hardware reset Xosera will typically reconfigure twice, but this is not a biggie (but 80ms of your life you may not get back each boot). 😅 With this change we don't really need UPduino HW reset hooked up (I suggest a jumper, perhaps).

### DIFF
--- a/code/firmware/rosco_m68k_development/videoXosera/xvidcon.asm
+++ b/code/firmware/rosco_m68k_development/videoXosera/xvidcon.asm
@@ -48,21 +48,6 @@ HAVE_XOSERA::
     rts
 
 
-; Delay for D0.w loops...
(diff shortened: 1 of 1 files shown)


roscopeco (approved):
Looks good, tests fine locally, thanks @XarkLabs :)

PR #229, opened 2021-06-29
Switch to storing SR rather than blindly enabling interrupts

roscopeco:


roscopeco:
Thanks for looking @0xTJ :) I've tested this with no video and with Xosera, but I don't have a working V9958 to test with at the moment :(

XarkLabs (approved):
Looks good, I can test with V9958 soon-ish.

PR #230, opened 2021-06-29
Add interrupt priority mask parameter to mcEnableInterrupts in machine lib

### LINKED ISSUE #226: mcEnableInterrupts in machine lib enables *all* interrupts
The `mcEnableInterrupts` function in the `machine` standard lib indiscriminately enables all interrupts (it sets the mask to 0).

This could potentially cause problems. A better solution would be to have the function take a new level mask as an argument.

0xTJ:
Adds a parameter to set interrupt priority mask in `mcEnableInterrupts`. Also, `mcDisableInterrupts` returns the interrupt priority mask as it was, to allow restoring the existing level if desired.

Fixes #226


roscopeco (approved):
Looks great, thanks @0xTJ  :)

PR #231, opened 2021-07-01
Lots of tidying up for FW2.0 and beyond

roscopeco:
This is unfortunately quite a large PR, however a lot of is just tidying up the equates and switching to consts for SDB table entries etc. However, there is some real change hidden in here. Briefly, the changes are (bigger ones in **bold**):

* Rename equates to `public` and `private`, indicating what is (should be) safe for non-firmware code to use
* Move 68681 equates to public, to sit with 68901. I'm not sure I like this - perhaps these can all be made private soon.
* Tidy up bootstrap code (use equates rather than magic numbers for SDB locations)
* **Move red LED toggle** out of `start_serial` into firmware (in prep for removing MFP addresses from public later)
* **Remove debug_stub** as I think this is unnecessary (happy to discuss however if I've missed something).
* Update InterfaceReference with cpu info SDB field
* Add `CPU` option to `starter_c` and `starter_asm` makefiles in prep for the future...

Further to the removal of debug_stub, I'm aware that this isn't ideal for anyone on Firmware 1.3 (non-BIGROM) or below. However, they ought not to build from `develop` now anyway (as the libs are incompatible, for example). However, it would be nice to continue to support released versions via a path of least resistance - to this end, I propose we switch the default branch in the repo away from `develop` to either the latest release branch, or `master` (IMO the latter better as we won't need to keep changing it). Feedback and ideas welcome on that too :)


XarkLabs (approved):
Sounds good.  Maybe just move debug_stub to be an example?  AFAIK, no harm including it even if it is firmware (but this may change?).

PR #233, opened 2021-07-06
Add SST flash lib and demo

roscopeco:
Adds a library to support direct interface (including erase and write) to SST39SF0x0 flash devices, and a small demo for use with HUGEROM / Pro systems.

The included demo is also intended to serve as a useful POC for a writeable configuration area in ROM as suggested by @XarkLabs . 

Proof of life; Notice "old data" (read from flash) and "new data" (written to flash) is swapped with each run (note: powered down for 1m between the two runs):

Screenshot 2021-07-06 at 12 14 48


code/software/libs/src/sst_flash/sst.c
```diff
@@ -0,0 +1,139 @@
...
+
+    uint8_t last = (*p) & 0x40;
+    uint8_t this;
+
+    while ((this = ((*p) & 0x40)) != last) {
+        last = this;
+    }
+}
+
+static void unlock(uint32_t rom_base) {
```
0xTJ, line 42: From the datasheet, it seems to me that the DQ6 toggling only happens for the Program and Erase operations, not after individual writes of the commands. I'd think that we only want to call `wait` in `sst_sector_erase`, `sst_sector_erase`, and `sst_write_byte`, but not in `unlock` or in `sst_command`.

code/software/libs/src/sst_flash/sst.c
```diff
@@ -0,0 +1,139 @@
...
+#define SST_UNLOCK_A1       ((0x00005555<<1))
+#define SST_UNLOCK_D1       0xAA
+#define SST_UNLOCK_A2       ((0x00002AAA<<1))
+#define SST_UNLOCK_D2       0x55
+
+#define SST_SUBCMD_SECERASE 0x30
+#define SST_SUBCMD_CHPERASE 0x10
+
+static void write(uint32_t addr, uint8_t data) {
+    uint8_t *p = (uint8_t*)addr;
```
0xTJ, line 27: It's a minor thing, and I'm not sure it's needed, but I'd consider making all the `uint8_t *p` be `volatile`, just to make sure that the compiler doesn't have the chance to inline and optimize-out any writes at some point in the future.

0xTJ (commented):
Aside from those couple little things, the code looks good to me!

code/software/libs/src/sst_flash/sst.c
roscopeco: Interestingly, this was my read of the data sheet too, however in testing I found that, without the waits, it didn't work. I just tried it again and reliably get this:

```
Kernel received okay; Starting...


rosco_m68k Flash EEPROM identifier / write POC
**********************************************

EVEN ROM is  <Unknown>  [Manufacturer ???]
ODD  ROM is  <Unknown>  [Manufacturer ???]

Config area is 1 sector per ROM (2 total)
  Sector num : 0x7f
  Sector size: 4096 bytes
  Total size : 8192 bytes
  Base addr  : 0x00efe000

Old data: 0xaa55aa55 0xaa55aa55 0xaa55aa55...
New data: 0xaa55aa55 0xaa55aa55 0xaa55aa55...
```

_However_ , I now suspect that the call to write was just preventing an optimisation (which you mention in your other comment). Making the write value in `write` be volatile allows these `wait`s to be removed, while keeping everything working, so looks like this is a doubly-good shout, thanks! :)

code/software/libs/src/sst_flash/sst.c
roscopeco: See reply to your other comment, but long story short this is totally necessary, good spot!

With this being `volatile` and the `wait`s removed, I'm now getting this:

```
Kernel received okay; Starting...


rosco_m68k Flash EEPROM identifier / write POC
**********************************************

EVEN ROM is SST39SF040  [Manufacturer SST]
ODD  ROM is SST39SF040  [Manufacturer SST]

Config area is 1 sector per ROM (2 total)
  Sector num : 0x7f
  Sector size: 4096 bytes
  Total size : 8192 bytes
  Base addr  : 0x00efe000

Old data: 0x55aa55aa 0x55aa55aa 0x55aa55aa...
New data: 0xaa55aa55 0xaa55aa55 0xaa55aa55...
```

roscopeco (commented):
Thanks for looking @0xTJ , I've pushed some changes and added some comments for context :)

roscopeco:
I just pushed the first cut of a basic ROM reflash utility. Proof of life (notice different versions before and after):
Screenshot 2021-07-07 at 00 04 10

0xTJ (changes requested):

code/software/libs/src/sst_flash/sst.c
```diff
@@ -0,0 +1,136 @@
...
+
+static uint8_t read_manufacturer_id(uint32_t rom_base) {
+    uint8_t *p = (uint8_t*)rom_base;
+    return *p;
+}
+
+static uint8_t read_device_id(uint32_t rom_base) {
+    uint8_t *p = (uint8_t*)(rom_base | 2);
+    return *p;
+}
```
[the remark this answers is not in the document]
roscopeco: Good call, will do 👍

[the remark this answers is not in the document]
roscopeco: Done :)

0xTJ (approved):
I'm happy with the way things look here, and I'm able to build and run till expected failure on my end. :)

PR #235, opened 2021-07-11
Feature/update libs for hugerom

roscopeco:
Updating the libs for hugerom support.

* Adds a link script for executables that with correct values for hugerom
* Fixes a bunch of missing things in the regular link script too
* Fixes memcheck to work with hugerom

### DIFF
--- a/code/software/libs/src/start_serial/link_scripts/hugerom_rosco_m68k_program.ld
+++ b/code/software/libs/src/start_serial/link_scripts/hugerom_rosco_m68k_program.ld
@@ -81,8 +81,8 @@ PROVIDE(_EFP_ATA_WRITE  = 0x00000488);  /* ATA write                    */
 PROVIDE(_EFP_ATA_IDENT  = 0x0000048C);  /* ATA identify                 */
 
 /* ROM absolute addresses */
-PROVIDE(_FIRMWARE       = 0x00FC0000);  /* firmware address             */
-PROVIDE(_FIRMWARE_REV   = 0x00FC0400);  /* firmware revision code       */
+PROVIDE(_FIRMWARE       = 0x00E00000);  /* firmware address             */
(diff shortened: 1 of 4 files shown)


code/software/libs/src/start_serial/link_scripts/hugerom_rosco_m68k_program.ld
```diff
@@ -0,0 +1,135 @@
...
+ENTRY(START)
+
+ram  = 0x00002000;          /* start of user memory                     */
+
+MEMORY
+{
+  RAM  : org = ram,  l = 0x00f7e000     /* program can use all RAM > 8K */
+}
+
+/* program configuration symbols that may be useful to override */
```
0xTJ, line 27: Any thoughts on splitting out the ROM-independent code from these, and including that with an `INCLUDE`? It might make it easier to keep everything in sync.

XarkLabs: The linker symbols have the advantage that they are really the only way to have absolute addresses from C (with simple extern).  Also handy from asm, but less of an issue there.  It might be nice to have the firmware revision in the linker ld name though (since it is a "matched set").

XarkLabs (approved):
Seems like this is fine, extending the current stuff in a straightforward way.

code/software/libs/src/start_serial/link_scripts/hugerom_rosco_m68k_program.ld
roscopeco: This is probably a good idea @0xTJ , I guess we could move the common stuff out to a generic link script that then gets INCLUDED in the specific ones (which could then have the revision in the filename too as @XarkLabs suggests). 

Worth looking into I think. Would you mind raising an issue for it?

PR #236, opened 2021-07-11
Fix #194

### LINKED ISSUE #194: chmod commands in Makefiles break the build under WSL
It has been noted in conversation on Discord that `chmod` commands used in various makefiles do not work on Windows under WSL(2) , and will break the build. This is used in many Makefiles in the project, and should be fixed everywhere.

@XarkLabs makes the point that:

> Another option is to precede the command in the Makefile with a - and Make will ignore any error returned (e.g., -chmod $(foobar)). (That is harmless on Linux/Mac so seems a good fix - offhand without trying it).

We should:

* Try it
* If it works, do it:
  * In the firmware
  * In the examples
  * In the examples in `hardware_projects`
  * Anywhere else it may be lurking
* If not, investigate further and fix
  * Everywhere mentioned above

(Thanks to Jayney and PaulaM on Discord for reporting)

roscopeco:
Fix #194 (whereby chmod fails and breaks WSL builds).

### DIFF
--- a/code/firmware/old-firmware/Makefile
+++ b/code/firmware/old-firmware/Makefile
@@ -37,7 +37,7 @@ OBJECTS=bootstrap.o kmain.o api/api.o rtlsupport.o 3rdparty/printf.o \
 
 $(BINARY) : $(OBJECTS)
 	$(LD) $(LDFLAGS) $^ -o $@
-	chmod a-x $@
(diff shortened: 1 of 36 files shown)


0xTJ (approved):
Simple change, sounds like it's effective

PR #237, opened 2021-07-12
Re-add equates.S for building older firmware

roscopeco:
Re-introducing the old `shared/equates.S` to allow FW1.3 and lower to still be built from develop. 

Added a note that the file is frozen, and pointing to the new one instead for future updates.


XarkLabs (approved):
Sounds good.

PR #238, opened 2021-07-15
Band-aid for lib build order (again).

roscopeco:
Will look at a proper fix for this going forward. I suspect it's because
we're misusing included partial makefiles a bit.


XarkLabs (approved):
This seems to fix it for me. 🙂

PR #239, opened 2021-07-17
Move starter projects and use env var for build

roscopeco:
This change makes the starter projects use the `ROSCO_M68K_DIR` environment variable as introduced by @XarkLabs . This should make building projects based on them **much** easier and save people from having to pollute their repo directory with their own code.

I have also moved the starters into their own tree (alongside `software` rather than within it) - I think this will make them easier to find, and also has the benefit of not requiring the environment variable to not be set just to build the example code (e.g. for people who are just starting out, this might be a nice thing to have).


0xTJ (approved):
Looks good, and copying to build out-of-tree worked for me. :)

PR #241, opened 2021-07-25
Direct EASy68K `SEND_CHAR` to console, not UART

### LINKED ISSUE #240: TRAP #15 Function #6 should print, not send (despite the name)
The Easy68k TRAP #15 function #6 , despite being named `SENDCHAR`, is documented to _display_ the character, so it should probably use the underlying `PRINTCHAR` EFP rather than the `SENDCHAR` one.

It might be worth (somehow) checking we're aligned with Easy68k on this...

(See: https://github.com/rosco-m68k/rosco_m68k/blob/develop/code/firmware/rosco_m68k_v1.3/InterfaceReference.md#1327-sendchar-function-6 )

0xTJ:
`TRAP #15` function `#6` now calls `FW_PRINTCHAR` instead of directly calling `SENDCHAR`. Previously, this one specific function wouldn't print to an attached display, instead always printing to the default UART. It will now print to the default console, matching functions `#0`, `#1`, `#3`, etc.

Fixes #240

### DIFF
--- a/code/firmware/rosco_m68k_development/easy68k/syscalls_asm.asm
+++ b/code/firmware/rosco_m68k_development/easy68k/syscalls_asm.asm
@@ -239,7 +239,7 @@ READ_CHAR:
 * ************************************************************************** *
 SEND_CHAR:
     move.l  D1,D0                       ; Platform code expects arg in D0...
-    bsr.w   SENDCHAR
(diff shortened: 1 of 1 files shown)


roscopeco (approved):
Awesome, thanks @0xTJ  ! Looks great, tested locally and works well 👍

PR #242, opened 2021-07-25
Add version info and flag check to flash update utility

roscopeco:
This change adds a new flag to the version flags in the firmware to indicate that the firmware is built as a (flashable) HUGEROM. Adding this after forgetting to build with HUGEROM one too many times, and flashing ROMs that wouldn't work.
A check is added to the update flash utility to check for this flag.

* Add version flag for HUGEROM builds
* Correctly set "Requires extended data area" flag in development ROMs
* Add `RomVersionInfo` to `machine.h` to allow versions to be checked easily
* Add check in `updateflash` to check version in the loaded ROM before flashing.


0xTJ (approved):
Looks good! Tested locally, and it correctly accepts/rejects builds.

Rather coincidental timing, I made this mistake this morning. Twice. 😅

PR #249, opened 2021-09-09
Patch Xosera console for new reg layout

roscopeco:
Temporarily patching Xosera console for new register layout.

This is only intended to keep us rolling until new and improved terminal comes along :)


roscopeco:
Closing, superseded.

PR #250, opened 2021-09-17
Add NO_TICK option to disable MFP interrupts

roscopeco:
For advanced testing (currently with Xosera), this adds an option to completely disable the system timer tick (and thus, MFP interrupts under firmware control).

This is primarily to work around the known issue with MFP IACK on r1.x boards, but may be generally useful for other things too.


0xTJ (approved):
Looks good 👍 there is/isn't tick when appropriate

PR #254, opened 2021-09-22
Remove libs `mno-align-int` and `mno-strict-align` options

### LINKED ISSUE #253: Standard libs appear to be causing an address error
This issue can be replicated by using the `fill` command in the `vterm` example program with FW 1.3. 

When the `fill` command is executed, the board bombs out with an address error. With the debug handler in the firmware, this reports as:

```
You can type with a working backspace, or use the arrow keys to move around.
Available commands: clear, color4, color8, trucolor, invert, bold, underline, normal, bell, blink, fill, cursor-pos, screen-size, help
$ 
  Unknown command: ""
$ fill
      Enter a character to fill the screen with: 

*** Software Failure: Address error (fault addr 000FFF3B)
PC=0000432A op=2F7C     SR=2000    USP=7F7F6FFF
d0=00000032 d1=0000000A d2=00000032 d3=0000000A
d4=00000010 d5=000F0000 d6=F7FF5FDF d7=FFFE7BFF
a0=00006938 a1=0000682E a2=00006938 a3=000069A8
a4=00006A1C a5=00005AE4 a6=000FFFF8 a7=000FFEDA
```

A glance at the `.map` file for vterm suggests this is in `__utoa` in `libcstdlib.a` .

0xTJ:
Removes the `mno-align-int` and `mno-strict-align` options from the libraries Makefile. 

Taking a look at the GCC source, `mno-align-int` (16-bit alignment) is already the default case, so it wasn't doing anything. While `malign-int` does improve performance on 32-bit-bus models, it's documented as an ABI-breaking option since it changes the usual structure alignment/padding.

`mno-strict-align` ignores the alignment requirements, and causes the crashes observed in #253. In all types except Coldfire, `mstrict-align` is the default, and is what we want.

For non-crashing ABI-conforming code, we don't want either of these options.

Fixes #253

### DIFF
--- a/code/software/libs/Makefile
+++ b/code/software/libs/Makefile
@@ -9,7 +9,7 @@ DEFINES:=$(DEFINES) -DROSCO_M68K_SYSLIBS
 CFLAGS=-std=c11 -ffreestanding -ffunction-sections -fdata-sections \
 				-nostartfiles -Wall -pedantic -Werror              \
 				-mcpu=$(CPU) -march=$(CPU) -mtune=$(CPU)           \
-				-mno-align-int -mno-strict-align $(DEFINES)
+				$(DEFINES)
 ASFLAGS=-Felf -m$(CPU) -quiet $(DEFINES)
 ARFLAGS=
 CC=m68k-elf-gcc


roscopeco (approved):
Looks good, thanks @0xTJ 👍

PR #255, opened 2021-09-24
Feature/printf fmt warn

XarkLabs:
This adds GCC attributes to the header with printf and related functions to cause them to type check arguments and issue a warning on type mismatches.  This can be a bit a picky, but usually can be made happy with small tweak to format (and that is worth it for finding major mixu-ps).  This PR also fixes any new warnings in rosco_m68k software.


XarkLabs:
Not this also includes updated Xosera register definitions (accidentally), but mostly useful. 🙂

roscopeco (approved):
Looks great, thanks @XarkLabs !

PR #263, opened 2021-10-20
Archive old firmware

roscopeco:
Just moving old firmwares to an `attic` directory.


roscopeco:
> Is having the Makefiles be copied (as far as I can tell) instead of moved intentional? This leaves the directories out of the `attic`, containing just the Makefiles.

Good spot, thanks :) I literally have no idea how that happened, but I think it's fixed now 😅

0xTJ:
Looks good, nice to clean up old stuff and get it out of the way.

PR #266, opened 2021-11-11
Support C++ (and C with GCC extension) global ctors/dtors

roscopeco:
This PR adds the most basic support for C++ user code (basically, removing the global constructor/destructor footgun by supporting the basics of the Itanium CXX ABI as used by GCC-m68k: https://itanium-cxx-abi.github.io/cxx-abi/abi.html#dso-dtor-runtime-api).

It also adds basic support for `__cxa_atexit` which G++ uses to register C++ destructors. This is not (currently) exposed to user code, but could quite easily be used to support user-code `atexit` handlers if desired.

There is a new test (in `code/software/tests/cpptest`) that tests the new functionality. Specifically, it exercises:

* C++ global constructors / destructors
* GCC `__attribute__((constructor))` and `__attribute__((destructor))` functions
  * With and without explicit priority

I have also run tests locally that check this doesn't break any existing code (that does not make use of these features) and am satisfied it doesn't.

Proof of life / Works On My Machine™:

Screenshot 2021-11-11 at 20 58 13

### DIFF
--- a/code/software/libs/src/start_serial/init.S
+++ b/code/software/libs/src/start_serial/init.S
@@ -68,7 +68,7 @@ PREMAIN:
     bsr.s   CALL_CTORS            ; Call global constructors
     lea.l   kmain,A0
     jsr     (A0)                  ; Fly user program, Fly!
-    move.l  #0,-(A7)              ; Call __cxa_atexit with NULL dso_handle
+    move.l  #0,-(A7)              ; Call __cxa_finalize with NULL dso_handle
     lea.l   __cxa_finalize,A0
     jsr     (A0)
     add.l   #4,A7
--- a/code/software/libs/src/start_serial/init.S
+++ b/code/software/libs/src/start_serial/init.S
@@ -3,10 +3,10 @@
 ;  ___ ___ ___ ___ ___       _____|  _| . | |_
 ; |  _| . |_ -|  _| . |     |     | . | . | '_|
 ; |_| |___|___|___|___|_____|_|_|_|___|___|_,_|
-;                     |_____|    firmware v1.2
+;                     |_____|           stdlibs
 ;------------------------------------------------------------
-; Copyright (c)2020 Ross Bamford See top-level LICENSE.md for licence
-; information.
+; Copyright (c)2020-2021 Ross Bamford See top-level LICENSE.md 
+; for licence information.
 ;
 ; This is the initialization code. The loader jumps into this code after the
 ; "program" is received via serial.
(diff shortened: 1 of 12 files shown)


0xTJ:
Very cool work! It's nice to see it working! One little comment I have is that currently `exit()` just warm-boots, while *I think* it should call the destructors. Is that something you think should be added, or are good without for now?

code/software/libs/src/start_serial/init.S
```diff
@@ -60,9 +63,57 @@ POSTINIT:                         ; running from copied run addr location
...
 
 PREMAIN:
-    move.l  $0004,-(A7)           ; push (soft) reset vector if kmain returns
     lea.l   __kinit,A0
     jsr     (A0)                  ; prepare C environment
+    bsr.s   CALL_CTORS            ; Call global constructors
     lea.l   kmain,A0
-    jmp     (A0)                  ; Fly user program, Fly!
+    jsr     (A0)                  ; Fly user program, Fly!
+    move.l  #0,-(A7)              ; Call __cxa_atexit with NULL dso_handle
```
0xTJ, line 71: A very minor thing, but I think this comment should read `__cxa_finalize`.

roscopeco: > Very cool work! It's nice to see it working! One little comment I have is that currently `exit()` just warm-boots, while _I think_ it should call the destructors. Is that something you think should be added, or are good without for now?

Ah yes, excellent point, good spot. I'll look into a way to make this happen cleanly 👍

roscopeco: > A very minor thing, but I think this comment should read `__cxa_finalize`.

Yup, it should. Will fix, thanks :)

roscopeco: Having looked into `exit` calling destructors, I've found that it's a bit tricky. For that reason, I'm proposing that we wait until `r2-initial-import` is merged before fixing (and accept that it will not work as planned with this merge - to be fair, it's already broken with HUGEROM, as noted at the bottom of the ticket linked below).

I've raised https://github.com/rosco-m68k/rosco_m68k/issues/269 to cover fixing this properly (post this merge and `r2-initial-import`) and will do the work as soon as the other branch is merged.

Let me know (or just resolve this conversation) if that's acceptable. Or if you can think of a better way to solve it, let me know :)

roscopeco: (I fixed the easy comment issue btw :sweat_smile: )

0xTJ: Makes sense to me!

0xTJ (approved):
Looks good to me, replicated results as expected. :)

PR #267, opened 2021-11-12
Add uberlib

roscopeco:
This PR adds a single static library archive containing all rosco_m68k libraries in one dependency.

The intention is to simplify user code build by allowing just `-lrosco_m68k` to be linked rather than requiring the individual libs to be supplied in the correct (and increasingly-complex) order.

Thanks to @XarkLabs 's work to implement `--gc-sections` properly in the libs, this will have a negligible effect on user code (as unused stuff will simply be dropped by the linker). It **will** have  small effect on some things though, as `printf-softfloat` is the default - I think with the baseline configuration now having 115.2kbps and SD card, this is an acceptable increase. Of course is size is a major concern the individual libs are still available and can be linked as before.

(It's worth noting that, depending on what is used from the standard libs, `-lgcc` is likely to still be required - this is expected).

### DIFF
--- a/code/software/libs/Makefile
+++ b/code/software/libs/Makefile
@@ -26,6 +26,8 @@ INSTALL_DIR=./build
 INSTALL_LIB=$(INSTALL_DIR)/lib
 INSTALL_INC=$(INSTALL_DIR)/include
 
+UBERLIB=librosco_m68k.a
+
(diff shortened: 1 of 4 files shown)


0xTJ (approved):
Tested and looks good to me!

roscopeco:
> Tested and looks good to me!

Thanks @0xTJ :) Will merge.

Issue #264, opened 2021-10-23
EhBASIC Makefile and build instructions need updating...


XarkLabs:
EhBASIC Makefile and build instructions need a bit of updating.  Currently I believe the Makefile ignores the MC68681 option and always tries to use the FIRMWARE_IO.  This is a problem for older firmware, so Makefile should probably default to new firmware, but still honor other options (my bad).

PR #268, opened 2021-11-12
Feature/fix 264

roscopeco:
Fixing https://github.com/rosco-m68k/rosco_m68k/issues/264

(Also removes an old archive that probably shouldn't have been committed in the first place 🤷 ).


0xTJ (approved):
Simple change, looks good to me! (Though I haven't tested if it behaves properly on older firmware.)

roscopeco:
> Simple change, looks good to me! (Though I haven't tested if it behaves properly on older firmware.)

Thanks @0xTJ , I believe it works on old firmware. Will merge.

PR #270, opened 2021-11-13
Bring in Shmall as a simple heap allocator

roscopeco:
This PR brings in Shmall (https://github.com/CCareaga/heap_allocator#shmall---simple-heap-memory-allocator) as a simple heap allocator for the standard libs.

* Bring in Shmall as a standard lib (`-lheap`)
* Use Shmall as a basic heap allocator for `malloc` and `free`.
* Use Shmall as a basic heap allocator for C++ `new` and `delete` (requires include of `heap.h`)
* Initialize standard heap in a global constructor, allowing 32KB for stack


0xTJ:
Is requiring `heap.h` to be included a design choice, or just a result of limitations in how things are set up for it? By moving the operators to `rosco_m68k.cpp` (renamed from `rosco_m68k.c`, adding `extern "C"` guards as appropriate) and adding C++ support to the `libs` `Makefile`, I can build `malloc-test` without including `heap.h`. This does in theory slightly reduce performance because those wrappers aren't inlined, but it does avoid that extra header include.

If this is of interest, I can branch and PR into feature/add-shmall-as-library, otherwise I'm happy to approve this one.

roscopeco:
That sounds great @0xTJ , if you would raise a PR for that into this branch, that would be much appreciated :) Having to include `heap.h` was just a result of limitations, it'd be great not to have to have that 👍

0xTJ:
I have noticed one small issue, the `dhrystone` program uses `malloc`, but doesn't link `heap`. There might be others that have this problem, but I run into the `__kinit` issue before I get to them, if any.

roscopeco:
> I have noticed one small issue, the `dhrystone` program uses `malloc`, but doesn't link `heap`. There might be others that have this problem, but I run into the `__kinit` issue before I get to them, if any.

Good spot, I'll sort `dhrystone` out now, and take a peek at others while I'm at it.

roscopeco:
Been through and checked everything that builds, `dhrystone` and `vterm` were the only things that needed `heap` linking in. I've also added a (currently very simple & inefficient) `calloc` as this is needed by `vterm`.

0xTJ (approved):
Looks good to me! :)

PR #272, opened 2021-11-16
Remove the need to include <heap.h> in C++ code to use new/delete

0xTJ:
This PR makes `rosco_m68k.c` in the `shmall` library into a C++ source file. Most of the content is contained in `extern "C"`, but implementations of `new` and `delete` are added.

All overloads which are defined as implicitly declared in every translation unit, even with the `<new>` header isn't included, are defined, except for those that use `std::align_val_t` or `std::nothrow_t` arguments. The `std::size_t` argument types are also replaced with `size_t`. The overloads call each other as specified for a standard library implementation.

The `malloc-test` test builds and runs cleanly, even after the `<heap.h>` header has been removed.

### DIFF
--- a/code/software/libs/Makefile
+++ b/code/software/libs/Makefile
@@ -5,14 +5,18 @@
 
 CPU?=68010
 EXTRA_CFLAGS?=-g -O2
+EXTRA_CXXFLAGS?=-g -O2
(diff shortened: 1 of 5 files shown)


roscopeco (approved):
Looks good to me! Thanks @0xTJ 👍

PR #275, opened 2021-11-16
Feature/fix 274

### LINKED ISSUE #274: Memcheck / sdfat_menu broken
Since recent changes in the standard libs, `memcheck` and `sdfat_menu` are broken.

This is because they define their own `__kinit` function to support "private stack". This conflicts with the `__kinit` in the `start_serial` lib. 

@XarkLabs suggests the fix is to make the lib `__kinit` weak, using `__attribute__((weak))`.

roscopeco:
Fix #274. Thanks to @XarkLabs for suggesting the fix!


0xTJ (approved):
Looks good to me!

PR #280, opened 2021-12-02
Fix easy68k comment (config is three bytes)

roscopeco:
The byte that was `E68K_RESERVED` is now `INTFLAGS`...


0xTJ (approved):
Looks good :)

PR #281, opened 2021-12-06
Fix `calloc` and `realloc`

### LINKED ISSUE #273: `calloc` and `realloc` need implementing properly
With latest changes, `calloc` is a super-simple implementation that works, but is inefficient and slow. 

`realloc` is implemented "bare-minimum to spec" (i.e. it always fails, returning `NULL`).

Both of these need implementing properly (in `libs/src/shmall/rosco_m68k.cpp`).

0xTJ:
This implements `realloc`, and adds a platform-specific fast clear and fast copy for the `calloc` and `realloc`, respectively.

Also fixes the heap size being set to `HEAP_INIT_SIZE` instead of the actual space, and fixes faulting on freeing a `NULL` pointer.

I've added some benchmarks and tests to `malloc-test`. The benchmarks should print out a number of ticks, but I always end up with the difference being 0, it seems like the ticks stop (blinking stops) once the program is loaded. I'm not sure if there's some quirk somewhere or if I've got something setup wrong, but my benchmarks came from external timing instead.

I compare the non-platform-specific pure C default case to the case with `__m68k__` enabled (two data sizes) below:

| Benchmark | `calloc` | `realloc` |
| --- | --- | --- |
| Pure C | 84 s | 97 s |
| M68k fast word | 18 s | 25 s |
| M68k fast long | 13 s | 19 s |

Fixes #273.

### DIFF
--- a/code/software/tests/malloc-test/kmain.c
+++ b/code/software/tests/malloc-test/kmain.c
@@ -20,6 +20,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <heap.h>
+#include <machine.h>
 
 static const char *test = "TESTING!TESTING!TESTING!TESTING!";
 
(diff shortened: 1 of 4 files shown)


code/software/tests/malloc-test/kmain.c
```diff
@@ -20,24 +20,141 @@
...
+
+    if (success) {
+        printf("calloc zero check      : succeeded\n");
+    }
+}
+
+void calloc_benchmark(size_t count, size_t size) {
+    printf("calloc benchmark       : starting\n");
+    bool success = true;
+    unsigned start_ticks = _TIMER_100HZ;
```
roscopeco, line 63: Hmm, unless I'm missing something, I think you'd want this to be something like:

```
unsigned start_ticks = *(uint32_t*)&_TIMER_100HZ;
```

(I'll admit these linker `PROVIDE`s always cause me some thinking, but something like the above is what I've generally used in other code and found to work 🤔 )

code/software/tests/malloc-test/kmain.c
```diff
@@ -20,24 +20,141 @@
...
+        void *ptr = calloc(count, size);
+        if (!ptr) {
+            printf("calloc benchmark       : failed to calloc(0x%zX, 0x%zX)\n", count, size);
+            success = false;
+            break;
+        }
+        free(ptr);
+    }
+
+    unsigned end_ticks = _TIMER_100HZ;
```
roscopeco, line 75: (And of course similar to the above comment here)

roscopeco (commented):
This looks great @0xTJ , thanks :) Left a couple of comments (based on your comment about the tick not being updated - I think that may actually be two separate issues. The blinking LED I think I've addressed in another PR, while the tick not appearing updated I think will be addressed by my comments here :) ).

code/software/tests/malloc-test/kmain.c
0xTJ: Hmm, I'll have to test that out once I've got the system tick fix up and running. Though looking at `machine.h`, it looks like `_TIMER_100HZ` is already `extern volatile uint32_t`, so I'm not sure that'll change things, aside from casting away the `volatile`.

roscopeco: Hmm, yes, you're right. It should _just work_ . I just did a little test here, and it seems fine:

```
TEST(system_tick_is_running) {
    uint32_t start_ticks = _TIMER_100HZ;

    do {
        opt_guard++;
    } while (opt_guard < 1000000);

    uint32_t tt= _TIMER_100HZ;
    printf("_TIMER_100HZ = 0x%08lx\n", tt);

    ASSERT(start_ticks != _TIMER_100HZ);

    return true;
}
```

Gives:

```
_TIMER_100HZ = 0x00067266
 * system_tick_is_running - _TIMER_100HZ = 0x000673c4
PASSED
```

So please ignore my previous comment :D

0xTJ:
I've made an updated results table with the actual ticks values:

| Benchmark | `calloc` | `realloc` |
| --- | --- | --- |
| Pure C | 8059 ticks (100%) | 9330 ticks (100%) |
| M68k fast word | 1683 ticks (21 %) | 2338 ticks (25 %) |
| M68k fast long | 1180 ticks (15 %) | 1837 ticks (20 %) |

roscopeco (approved):
Looks great (and the numbers are very impressive)! Thanks @0xTJ ! :)

PR #283, opened 2021-12-07
Fix #282

### LINKED ISSUE #282: LED Blink on r2 board is unreliable
The interrupt-driven LED blink on the r2 board with firmware 2.0 is unreliable, and will sometimes stop blinking (especially noticeable when the computer is under load, such as when receiving via Kermit).

This appears to be a simple issue in the firmware. Specifically, the `TICK_HANDLER` in `bootstrap.asm` is writing the byte part of a register, but testing a word. So this code:

```
move.b  SDB_INTFLAGS,D1
tst     D1                        ; Is INTFLAGS zero?
```

should actually use an explicit size for the `tst` instruction (which otherwise defaults to `test.w`):

```
move.b  SDB_INTFLAGS,D1
tst.b   D1                        ; Is INTFLAGS zero?
```

I will push a branch to fix this issue shortly.

roscopeco:
Hopefully fixes #282. Seems to work here, but would appreciate if someone else could test (as the issue was intermittent anyway, I could just be getting "lucky").

@0xTJ I think this might address the lack of blinking LED you noted in your PR 🤔

### DIFF
--- a/code/firmware/rosco_m68k_v2.0/bootstrap.asm
+++ b/code/firmware/rosco_m68k_v2.0/bootstrap.asm
@@ -625,7 +625,7 @@ TICK_HANDLER:
     beq.s   .TICKRESET                ; bail now if not...
 
     move.b  SDB_INTFLAGS,D1
-    tst     D1                        ; Is INTFLAGS zero?
+    tst.b   D1                        ; Is INTFLAGS zero?
     beq.s   .TURNON                   ; If so, go to turn on
     
     ; If here, LED is already on, turn it off


0xTJ:
It's a good catch, though don't think this is the source of my particular issue, since I've still been using `REVISION1X` builds. In my case, the first half of the issue seems to be Kermit stopping the interrupts, since loading from an SD card actually seems to work, but then it fails once I get to a `realloc` part, so I've got to investigate that further.

roscopeco:
> It's a good catch, though don't think this is the source of my particular issue, since I've still been using `REVISION1X` builds. In my case, the first half of the issue seems to be Kermit stopping the interrupts, since loading from an SD card actually seems to work, but then it fails once I get to a `realloc` part, so I've got to investigate that further.

Ah right, I didn't realise you were still on `REVISION1X` - in that case, agreed, this isn't likely to be your issue.

I've verified your change (along with this one) works on r2, but I'll set up a r1 board in a bit and run your test here to see if I can replicate :)

0xTJ (approved):
Still not sure what the issue with my r1.23 board with `REVISION1X` `HUGEROM` 2.1.DEV is, but this fix seems effective on my now-running 2.1 board! The LED keeps on ticking, and I get good numbers on that `malloc-test`.

PR #287, opened 2021-12-12
Use DUART timer instead of counter

0xTJ:
This PR uses the DUART timer instead of counter mode of the C/T. This should keep the timing more accurate because counting resumes without having to reissue a start command. The datasheet recommends this as the mode for periodic interrupts.


roscopeco (approved):
Looks great, thanks for this @0xTJ !

PR #290, opened 2021-12-21
Make FW1.3 buildable with GCC 11

roscopeco:
* Taking the easy road, just cutting down a (rarely-seen) error message to shave some bytes...
* Tested with latest GCC from home-brew tap


0xTJ (approved):
Build succeeds here too. It's the cheap and easy fix, but it's good enough! 👍

PR #291, opened 2021-12-21
Feature/fix 285

roscopeco:
Make HUGEROM build be the default for all example code in `develop` now that r2 is out.


0xTJ (approved):
Makes sense to build for the latest. They seem to build as HUGEROM for me as expected. 👍

PR #293, opened 2022-01-10
Feature/release fw 2.1

roscopeco:
Release FW 2.1. This is a minor release:

* Fixes intermittent issue with blinking LED
* Use DUART timer mode
* Much better alloc support in libs
* Tidy ups


0xTJ:
Should the `code/firmware/rosco_m68k_v2.0` directory be renamed/copied to `code/firmware/rosco_m68k_v2.1` to keep the versions in sync?

0xTJ:
Also, it looks like there's an inconsistency with how versions are decoded. `RELEASE_VER` seems to use BCD, so this is version `0210`. `updateflash` instead uses the value directly, so it detects this ROM image as being version `2.16` instead of `2.10`.

roscopeco:
Good spot @0xTJ , done on both counts :)

0xTJ (approved):
Looks good to me 👍, most of the major changes are parts of previous PRs.

Those extra commits will probably need to be brought into develop too.

PR #296, opened 2022-01-12
Fix #269

### LINKED ISSUE #269: With latest changes, cstdlib exit() does not call destructors
With latest changes, the `exit` function in `cstdlib` does not correctly call destructors. It should.

This is complicated slightly because the `exit` function is defined in the `cstdlib` library, and there is no requirement that `start_serial` actually be in use in order to use that library. So it's possible that global destructors are not even _supported_ when `exit` is called. 

I propose the way to fix this is to add an EFP entry for e.g. `EFP_PROGRAM_EXIT`. The firmware can set this the same as the reset vector (remembering to update it to `warm_boot` in `main1.c`) and then `start_serial` can change it to the address of a label _after_ kmain returns in `init.S`.

However, this change is tricky currently (due to `r2-initial-import` being in-flight) so I am proposing we defer until after than merge (and then fix under this ticket, which I will assign to myself).

**Note**: While investigating, I found that `abort` (also in `cstdlib.c`) will need updating when r2-initial-import is merged, as it currently hardcodes the ROM start address in inline assembly. This will instead need to be taken from the linker-provided symbol.

roscopeco:
Fix #269 by:

* Introducing new `EFP_PROG_EXIT` at `0x490`.
* Initially set new EFP to reset vector (and update when warm boot is setup)
* In `start_serial` lib `init.S`, save that EFP and swap it out for a label just after main.
* When we hit that label (either through jump to EFP or normal main return), restore saved EFP
* Update libs `exit` function to jump to EFP

This PR also fixes an hardcoded ROM address in `abort`, changing it to the reset vector in RAM.

### DIFF
--- a/code/firmware/rosco_m68k_v2.1/InterfaceReference.md
+++ b/code/firmware/rosco_m68k_v2.1/InterfaceReference.md
@@ -1372,7 +1372,7 @@ must be accessed through the public TRAP functions!
 | 0x438   | FW_CLRSCR - Clear the default console (where supported)                                           |
 | 0x43C   | FW_MOVEXY - Move cursor to (X,Y) (see note 1)                                                     |
 | 0x440   | FW_SETCURSOR - Show (D0.B > 0) or hide (D0.B == 0) the cursor                                     |
-| 0x444   | RESERVED                                                                                          |
+| 0x444   | FW_CHECKCHAR - Determine whether a character is available on the default UART                     |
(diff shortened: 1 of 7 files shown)


roscopeco:
> It might be a good idea document the new vector in `InterfaceReference.md`. Otherwise, the changes seem good!

Good call @0xTJ , thanks. I added the new vector to the reference.

0xTJ (approved):

0xTJ (approved):
Looks good to me! I get the destructors called as expected. 👍

PR #297, opened 2022-01-14
Fix #288

### LINKED ISSUE #288: `REVISION1X` and `HUGEROM` together don't set `ROMDEVICE` correctly
When building FW2.0 for `REVISION1X` with `HUGEROM`, the ROMs are built correctly for the larger ROM, but the `ROMDEVICE` setting is forced back to `AT28C256`, so `make clean burn` cannot be used.

Non-critical as there is a two-step workaround:

```
REVISION1X=true HUGEROM=true make clean all
ROMDEVICE=SST39SF040 ./burn.sh
```

But still, this should be fixed.

roscopeco:
Fix #288 by tidying up and centralising all the size-specific options into one place in the Makefile.


0xTJ (approved):
Looks good!

PR #300, opened 2022-01-19
Fix #284

### LINKED ISSUE #284: CPU detection does not work in a "fully-loaded" system
The current firmware relies on at least one bus error during early initialisation in order to detect the CPU model, which is then used to determine to multiplier for speed calculation. 

On r2 boards, there are three devices that have the potential to cause bus error during init - the 68681 (which _should_ always be present), Xosera and the V9958.

Assuming the 68681 will never cause a bus error, if a Xosera board is present (thus not causing a bus error) then the V9958 init is skipped completely (removing the last opportunity for a bus error). So in this configuration, the CPU model is never detected, and the firmware chooses a (rather optimistic) default:

IMG_0178

Suggestions for a fix (**UPDATE** Illegal Instruction trap decided as the way to go, see discussion in comments):

* ~Always generate a bus error (perhaps we can reserve a word in IO space that should always be unused, for example)~
* ~Stop relying on bus errors for CPU detection (we do this because the CPU stacks enough useful info for the handler)~
* Something else? **Yes - probe with CPU-specific instructions & illegal instruction trap**

roscopeco:
Fixes #284 using @0xTJ 's suggestion to use illegal instruction traps rather than parsing bus error frames.

* Adds `cputype.asm` to firmware, with routine called from `INITSDB` to init CPU type field
* Removes logic to determine CPU type from `BERR_HANDLER`
* Adds `cputype` as a user program (in tests) (I'm thinking we might want to build it out to support e.g. coprocessors)
* Fixes a couple of small Makefile issues
  * Squawk CPU type when building firmware
  * Show correct ROM device type at end of build

Testing:
* Tested with: 68010, 68020, 68030 (on Pro prototype)

It's worth noting that:

* This currently does more detection that we'll ever need on Classic (e.g. it'll never support 68040/68060). At some point though this might get unified with Pro firmware, so I'm keeping those in here for now.
* Currently, it's using `PMOVE` rather than `MOVE16` to detect 68030 vs 68040. This is because, on the pro during testing, MOVE16 was hanging the board. I suspect this is because the hardware doesn't correctly respond to CPU space coprocessor access, and so hangs the board rather than raising F-line. This is reflected in a comment, with the (future) code ready to be switched to when I get it working.


0xTJ (approved):
Looks great! 👍

PR #301, opened 2022-01-21
Feature/fix firmware rev display in ansiterm

roscopeco:
Recent change in the way firmware encodes version broke Xosera ansiterm (my bad). This should fix it.

Proof of life:
IMG_0225


0xTJ (approved):
Seems to fix it perfectly! :+1

roscopeco:
> Seems to fix it perfectly! 👍

Thanks for looking @0xTJ :)

PR #302, opened 2022-01-22
Don't try to return from a bus error on a 68000

0xTJ:
On Discord, aowron made me realize that the temporary bus error handler will try to `RTE` on a bus error when an MC68000 is used. This will fail because the stack layout is different than what the '000 `RTE` expects, and will probably try to jump to the address of the bus error.

Even though the MC68000 isn't really supported, it would be better to try to signal a bus error rather than to mysteriously crash (assuming that the frame isn't mistaken for that of another chip, but there's nothing we can do about that anyways).

### DIFF
--- a/code/firmware/rosco_m68k_v2.1/bootstrap.asm
+++ b/code/firmware/rosco_m68k_v2.1/bootstrap.asm
@@ -493,7 +493,9 @@ BERR_HANDLER::
     beq.w   .IS020 
 
     ; If we're here, assume it's a 68000.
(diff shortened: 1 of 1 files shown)


roscopeco (approved):
Looks good to me, and works well in local testing with 68000. 

I'd _like_ to find a way to make this work on 68000 (possibly by having the code that causes the fault stash a continuation address in `A6` or something) but that's likely to be "deeper surgery" and this is a good change for the interim, thanks @0xTJ :)

PR #304, opened 2022-01-27
Benchmark classic v2 68020 20MHz

roscopeco:
Classic v2 works with 68020 adapter at 20MHz (assuming compatible CPU of course) so it seems reasonable to add that config to dhrystone `bench.txt`.

The proof:
Screenshot 2022-01-27 at 16 03 00


0xTJ (approved):
Nice! That makes for a speedy computer. 👍

roscopeco:
Thanks for checking @0xTJ 👍

Issue #303, opened 2022-01-25
Warm boot doesn't re-initialise DUART after hardware reset


0xTJ:
Currently, the line https://github.com/rosco-m68k/rosco_m68k/blob/d2e599dbe13a67ec5b3cbc58612797911bdc1022/code/firmware/rosco_m68k_v2.1/main1.c#L116 has the warm boot handler iset to be used both at program exit and as hardware reset vector. This causes problems when using the system `/RESET` signal to reset the processor, because this also resets the DUART (and could reset any future added peripherals). The warm boot isn't made to re-setup the entire system, just to call the current program loader, so the warm boot message isn't printed, and Kermit doesn't work.

I'd propose that the hardware reset vector use a separate routine that's one step "colder" than the current warm boot.

It might also be worth considering splitting out some of the code in `bootstrap.asm` into separate files to make things a little more modular, but that's a separate issue.

roscopeco:
Interesting. This makes sense that it should be a problem, but I don't see it in (e.g. `updateflash` which uses the `reset` instruction and AIUI _should_ trigger this issue, but doesn't, at least on my system) 🤔 

Is there a test we can run which exhibits this behaviour?

(And yeah, I agree on `bootstrap.asm`. Breaking it up a bit has been on my list for quite a while now...)

roscopeco:
Having drunk plenty of coffee and thought more about this, I think it's good as currently implemented. I kind of feel like, if a piece of code is going to run a `reset` instruction (which it's worth noting doesn't actually reset the processor), then it's not really doing a warm reboot, and has no business expecting to be able to jump to the warm boot vector, as it has deliberately reset all the peripherals.

I think where `reset` is used, it's a reasonable requirement that the next thing the software (or firmware, I'm not against adding a trap for this) does is load the stack and initial PC from the ROM base and jumps there for a "cold" reboot.

What do you think?

roscopeco:
Ah right, okay, I can see your use-case now. 

I'm not _against_ the above in principle (but I do want to think it through more and digest it somewhat), and you're right that it's not technically supported with the current expansion bus (though likely just yanking `/RESET` will "work fine" with any 68010 you care to try it with). If you have chance to work something up I'd gladly look at it 😃 

FWIW I _do_ think we can achieve broadly what you need with a change in the glue logic. Specifically, if you replace IC5 with:

```
GAL22V10
GLUE

[...]

PR #305, opened 2022-01-28
Prototype system initialization changes for comment

0xTJ:
This is my prototype for comment, as discussed in #303. I've cut back the scope of my changes a bit.

In the end, most of my changes ended up preserving the original code, just reorganizing it. I've split out some code from `bootstrap.asm`, so that file is a good bit simpler. It has mostly ended up in the revision- or UART-specific files. While some of this technically repeats some bits of codes within routines, a lot of them were a bit hard to read with the amount of `ifd`s in them.

One major change is that the EFPT is completely initialized with dummy pointers before calling any system init functions, to make it easier to handle no default console being present.

I've tried to make the serial chips' code a little bit more orthogonal to the board revisions, though they're still tightly coupled.

I had to add `-fno-delete-null-pointer-checks` to the `Makefile`, because GCC would generate a `trap #7` when it could prove that a certain access would always dereference `NULL` (this happened when I changed a global pointer with address 0 to be `const`).

I added a `hot_boot` routine, which is like `warm_boot`, but doesn't print the banner. It's called from `main1` to make the state consistent with a warm boot when calling the program loader.

Actually running it, the changes to code flow are minor, and I don't think there would be any noticeable changes (at least in theory, practice might be different).

I haven't checked that it runs on r1.x hardware (I currently only have one MC68010, and it's a pain of a chip to transplant), just that it builds for `REVISION1X=true HUGEROM=false`.

I'd appreciate any comments on this. I've tried to be good about committing per-change, so it shouldn't be too hard to undo specific changes.

This is inspired by #303, but that proposed PLD change would close that issue.


roscopeco:
Apologies for the delay on this @0xTJ , I have been a bit short on time and want to pull the changes and test / look at them locally. I should be able to do that tomorrow 😃

code/firmware/rosco_m68k_v2.1/bootstrap.asm
```diff
@@ -61,51 +60,55 @@ VECTORS:
...
+    ; Copy exception vectors table to RAM at VECTORS_LOAD (0x00000000).
+    ; VBR defaults to that location anyway for 68000 compatibility.
+    lea     (VECTORS),A0                ; Vectors in ROM into A0 (source)
+    lea     (VECTORS_LOAD),A1           ; Vectors in RAM into A1 (destination)
+    move.l  #VECTORS_COUNT,D0           ; Count into D0 (DBcc only uses word size)
+    bra     .ISR_COPY_START             ; Jump to DBcc to start loop
+.ISR_COPY_LOOP:
+    move.l  (A0)+,(A1)+                 ; Copy long source to dest, with postincrement.
+.ISR_COPY_START:
+    dbf     D0,.ISR_COPY_LOOP           ; Decrement D0 and loop if not negative, ignore cc
```
roscopeco, line 82: Nice, much better 👍

code/firmware/rosco_m68k_v2.1/main1.c
```diff
@@ -71,10 +72,10 @@ noreturn void main1();
 
 noreturn void linit() {
   // copy .data
-  for (uint32_t *dst = &_data_start, *src = &_code_end; dst < &_data_end; *dst = *src, dst++, src++);
+  for (uint16_t *dst = &_data_start, *src = &_code_end; dst < &_data_end; *dst = *src, dst++, src++);
```
roscopeco, line 75: Interesting. Why this change?

code/firmware/rosco_m68k_v2.1/rosco_m68k_private.asm
```diff
@@ -20,12 +20,12 @@
 ; bit 14    : Requires larger system data area
 ; bit 15    : Snapshot version
   ifd HUGEROM
-RELEASE_VER     equ     $E0000220
+RELEASE_VER     equ     $E0000230
```
roscopeco, line 23: Did you mean to bump this? 2.20 is still in DEV, AFAIK this change should be part of that eventual release...

roscopeco (commented):
This looks generally good, and is a significant win in terms of clean separation and tidying up a bunch of things. It tests fine here on r2 boards in various HW configurations. I too have not had chance to test on an r1, but will do that next.

code/firmware/rosco_m68k_v2.1/rosco_m68k_private.asm
0xTJ: Very good point, I've reverted that.

code/firmware/rosco_m68k_v2.1/main1.c
roscopeco: Okay, yep, good points (and well spotted!). Given that I’d go with keeping this change 👍

roscopeco (approved):
Lightly tested on r1.23 board with AT28C256 ROMs (built with `REVISION1X=true`) and it works great. As mentioned, already tested on r2.x boards in various configurations.

Overall, I think this is a great change, much cleaner. Thanks again for the good work @0xTJ !

0xTJ:
Thanks for the review @roscopeco!

PR #306, opened 2022-01-28
External reset will enable `BOOT` line reset from ROM

### LINKED ISSUE #303: Warm boot doesn't re-initialise DUART after hardware reset

0xTJ:
This is a slight variant of the change proposed in https://github.com/rosco-m68k/rosco_m68k/issues/303#issuecomment-1021571118. Performing an external processor reset will enable the `BOOT` line, with countdown.

Initially, I used the one unused output as a feedback of `HWRST + /HALT * /RESET` as `DOBOOT`. This then took the place of `HWRST` in the registered equations. Afterwards, I noticed that I could keep the pin free by just using `/HALT` and `/RESET` directly, since `HWRST` drives those low.

This fixes #303, since external reset now resets the processor to ROM. It would be even more fixed if any future revisions had an optional header `/HALT` for a flying-lead off to something, like `CS2` has.

This seems to work properly for me, driving both `/RESET` and `/HALT` low properly resets the system. Interestingly, the `/RESET`-only reset that seemed to semi-work before didn't work once I disconnected the bus board (with Xosera and expansion RAM) from the mainboard, but using both signals still works properly.


roscopeco (approved):
Looks good, thanks @0xTJ  👍

PR #310, opened 2022-02-16
Bandaid fix for `<string.h>` in C++

### LINKED ISSUE #308: Error when including string.h in a cpp file
To reproduce, simply add a test.cpp with the content `#include <string.h>` to starter_c.

Output:
```
michi@michis-ibm-haupt:~/code/rosco_m68k/rosco_m68k/code/starter_projects/test_c$ make
m68k-elf-g++ -c -std=c++20 -fno-exceptions -fno-rtti -ffreestanding -ffunction-sections -fdata-sections -Wall -Wextra -Werror -Wno-unused-function -pedantic -I/home/michi/code/rosco_m68k/rosco_m68k/code/software/libs/build/include -mcpu=68010 -march=68010 -mtune=68010 -DROSCO_M68K  -o test.o test.cpp
In file included from test.cpp:1:
/home/michi/code/rosco_m68k/rosco_m68k/code/software/libs/build/include/string.h:23:35: error: ISO C++17 does not allow 'register' storage class specifier [-Werror=register]
   23 | char* memchr(register const char* src_void, int c, size_t length);
      |    
```

```
which m68k-elf-g++
/home/linuxbrew/.linuxbrew/bin/m68k-elf-g++
```

```
m68k-elf-g++ --version
m68k-elf-g++ (GCC) 10.2.0
```

I assume that `register` was an optimization.

0xTJ:
This fixes #308, where the use of the `register` keyword caused `<string.h>` to not be usable in C++. There were also some functions in it that had the wrong types.

While I was at it, I split out the content that was in `<string.h>` that should be in `<ctype.h>` or `<strings.h>`.

This is really only a stopgap, the non-freestanding part of the stdlib is still missing bits, most of the `<c...>` C++-style headers aren't in, and there's missing `extern "C"`, but it should fix this specific issue for now.

### DIFF
--- a/code/software/adventure/adventure.c
+++ b/code/software/adventure/adventure.c
@@ -34,6 +34,7 @@
 #include <ctype.h>
 #include <stdio.h>
 #include <string.h>
+#include <strings.h>
 #include <stdlib.h>
 #ifdef __CC65__
 #include <conio.h>
--- a/code/software/libs/src/cstdlib/ctype.c
+++ b/code/software/libs/src/cstdlib/ctype.c
@@ -1,0 +1,68 @@
+/*
+ *------------------------------------------------------------
+ *                                  ___ ___ _   
+ *  ___ ___ ___ ___ ___       _____|  _| . | |_ 
+ * |  _| . |_ -|  _| . |     |     | . | . | '_|
+ * |_| |___|___|___|___|_____|_|_|_|___|___|_,_| 
+ *                     |_____|       firmware v1                 
+ * ------------------------------------------------------------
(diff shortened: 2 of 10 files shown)


roscopeco (approved):
Looks good to me! Thanks @0xTJ

0xTJ:
I did a quick check before merging and noticed that I had made a small mistake in one of the `ctype` functions, accepting a tab instead of a space as printable.

roscopeco (approved):
Looks good, and good spot. Thanks @0xTJ 👍

PR #311, opened 2022-02-27
set ROM_BIN path for Make to embed a ROM image

XarkLabs:
Added build time option to updateflash so that when ROM_BIN is set during Make, it will embed the ROM image in the updateflash binary (so SD card setup not needed to update flash).  Tested with SD and without SD, seems to work nicely.


roscopeco (approved):
Looks good to me, and some nice doc cleanup too 🙂 thanks @XarkLabs !

PR #312, opened 2022-03-02
Duart spi opt

XarkLabs:
I am pretty happy with this.  Tested read and write, works nicely and ~20KB/sec. spi_transfer is still TODO, but it can wait (not used for SD).


0xTJ:
Looks good as far as I can tell! :+1: It works, and I don't notice any issues with the code.

roscopeco (approved):
Great stuff, thanks @XarkLabs ! 😀

PR #317, opened 2022-04-02
Fix endless loop in memcheck with expansion RAM

### LINKED ISSUE #316: Memcheck hangs if memory expansion board is connected to a v2 based rosco_m68k board
Running the memcheck example will hang during RAM scan if the memory expansion board is connected to the v2 classic board. 

The hang happens once the program hits the memory mapped area. It will not fail if the jumpers of the memory board are set to 0b111, no memory is installed in the expansion board board, or the board is not connected. 

I observed this on a 2.10 classic running firmware 2.16.

0xTJ:
This is a minor change where the scanning pointer `current` wasn't advanced when checking expansion RAM, leading to an infinite loop where the same address is checked forever.

Fixes #316.

### DIFF
--- a/code/software/memcheck/memcheck.c
+++ b/code/software/memcheck/memcheck.c
@@ -180,11 +180,13 @@ static KRESULT build_memory_map(volatile MEMINFO *header) {
         }
       }
       current += 16; // skip by 64 bytes to speed scanning
-    } else if (!block_started) {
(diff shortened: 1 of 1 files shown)


roscopeco (approved):
Huh, weird that we didn't notice this before now 🤔 . But anyway, LGTM, thanks @0xTJ 👍

PR #319, opened 2022-04-02
update starter_projects make and add to top make

XarkLabs:
Minor tweak to sync up starter project Makefiles and add them to the "master make" and the SD card copy (so they stay in rotation begin "tested").


roscopeco (approved):
Looks good to me, thanks @XarkLabs 👍

PR #321, opened 2022-04-05
Fix #320

### LINKED ISSUE #320: Firmware 2.2 non-functional with r1.23 mainboards
When the 2.2 firmware is built for revision 1.x boards, it appears that it doesn't correctly set up MFP interrupts, so the tick never happens and the board hangs waiting for the CPU speed loop.

Specifically, when built with:

```
BIGROM=true REVISION1X=true make clean burn
```

The firmware hangs after printing the banner, before printing the CPU speed. If building with `NO_TICK`:

```
BIGROM=true REVISION1X=true NO_TICK=true make clean burn
```

then the firmware functions correctly (but does not display CPU speed or operate the tick, of course).

I have verified that both Firmware 2.0 and Firmware 2.1 function correctly on revision 1.23 boards when build with `BIGROM=true REVISION1X=true` which leads me to believe the issue is somewhere in this changeset: https://github.com/rosco-m68k/rosco_m68k/compare/v2.1...v2.2

roscopeco:
Fixes an issue whereby the system tick no longer worked on revision 1.x systems (#320).

The issue was that there was a bit of DUART code accidentally pulled into the new `rev1.asm` file. This was attempting to reset the DUART timer, and not clearing the ISR flag on the MFP leading to an apparent hang.

I have tested this change on both r1.23 board with MC68901 and AT28C256 ROMs and on r2.12 board with XR68C681 and SST39SF020 flash, and found it working on both. Would appreciate any further testing pre-merge.

### DIFF
--- a/code/firmware/rosco_m68k_firmware/rev1.asm
+++ b/code/firmware/rosco_m68k_firmware/rev1.asm
@@ -62,8 +62,7 @@ TICK_HANDLER::
     move.w  D0,SDB_TICKCNT            ; ... and write back to SDB
     move.l  (A7)+,D1                  ; Restore D1
 
-    move.b  R_STOPCNTCMD(A0),D0       ; Clear ISR[3]
-    move.l  (A7)+,A0                  ; Restore A0
+    move.b  #~$20,MFP_ISRB            ; Clear interrupt-in-service
     move.l  (A7)+,D0                  ; Restore D0
 
     rte                               ; We're done


0xTJ (approved):
Ah, looks like this is my bad. :sweat_smile: 
I've tested this on a r1.23 board (MC68901) with SST39SF040, with no other expansions, and it seems to work fine! :+1:

roscopeco:
Awesome, thanks for checking @0xTJ  :)

PR #322, opened 2022-04-06
Fix MAME ROM build for HUGEROM/Classic v2

roscopeco:
Just fixing MAME ROM build to correctly generate 1MB ROM image for upcoming classic v2 support in MAME.

Build with `MAME=true make clean all` or `MAME=true REVISION1X=true make clean all`.


0xTJ (approved):
Looks good to me! :+1: 

The only comment I have is should `all` include the `mame` target when built with `MAME=true`?

PR #324, opened 2022-04-12
Feature/classic 020 r1

roscopeco:


code/pld/r2/r2020_cpu_glue/ic8_cpu_glue.pld
```diff
@@ -0,0 +1,33 @@
...
+Q1.R = /Q1 *  Q0
+     +  Q1 * /Q0
+
+
+Q2.R =  Q2 * /Q1
+     +  Q2 * /Q0
+     + /Q2 *  Q1 *  Q0
+
+
+Q3.R =  Q3 * /Q2
```
0xTJ, line 19: One thing that's probably very minor, is the E clock being low for 8 clock cycles, and high for 8, instead of low for 6, high for 4. I might take a crack at the PLD equations myself later to avoid adding too much work over such a small thing.

0xTJ (changes requested):

0xTJ:
It would be nice if the DUART could live in a contiguous block of memory, hooked up to the highest byte and using a port size of 8, but unfortunately that would require more changes (I think the intention of this was keeping it more or less like the r2, with minimal changes to allow for the '020), as well as changes to code (with an extra board type to juggle).

code/pld/r2/r2020_cpu_glue/ic8_cpu_glue.pld
```diff
@@ -0,0 +1,33 @@
...
+     +  Q3 * /Q1
+     +  Q3 * /Q0
+     + /Q3 *  Q2 *  Q1 *  Q0
```
0xTJ, line 22: I haven't tested this on hardware, but I figured out these equation in a spreadsheet, and they at least build for this PLD. `QX` is a no-connect pin that pulses once `Q2`-`Q0` get to either 4 or 2 (2 less than the cycle counts for each "half"-cycle), depending on whether `Q3` is low or high. On the next clock, that triggers the toggling of `Q3`, and resets the counter in `Q0`-`Q2`.

```suggestion
Clock DS    SIZ0    SIZ1    A0    AS   NC    NC    NC   GND
/OE   NC    LDS     UDS     QX    Q3   Q2    Q1    Q0   VCC

Q0.R = /QX * /Q0


Q1.R = /QX * /Q1 *  Q0
     + /QX *  Q1 * /Q0


Q2.R = /QX *  Q2 * /Q1
     + /QX *  Q2 * /Q0
     + /QX * /Q2 *  Q1 *  Q0


Q3.R = /QX *  Q3
     +  QX * /Q3


QX.R = /Q3 *  Q2 * /Q1 * /Q0
     +  Q3 * /Q2 *  Q1 * /Q0
```

code/pld/r2/r2020_cpu_glue/ic8_cpu_glue.pld
0xTJ: Follow-up to this added as separate comment.

code/pld/r2/r2020_cpu_glue/ic8_cpu_glue.pld
roscopeco: This looks good to me @0xTJ , and works great, thanks! As you say, probably minor (especially with full 6800 bus not really being supported on 68020) but good to stay in-spec as far as possible :) 

Screenshot 2022-04-20 at 06 40 22

roscopeco:
> It would be nice if the DUART could live in a contiguous block of memory, hooked up to the highest byte and using a port size of 8, but unfortunately that would require more changes (I think the intention of this was keeping it more or less like the r2, with minimal changes to allow for the '020), as well as changes to code (with an extra board type to juggle), so it makes sense not to do that right away.

Yep, as you suggest, I wanted this board to be as software-compatible as possible without having to have a separate build profile to support.

code/pld/r2/r2020_cpu_glue/ic8_cpu_glue.pld
roscopeco: Changes made (full reply and proof-of-life in other comment), thanks @0xTJ 👍 :)

0xTJ (approved):
Looks good to me! Happy to approve :)

roscopeco:
> Looks good to me! Happy to approve :)

Thanks for the suggestions @0xTJ and for taking another look 🙂

PR #327, opened 2022-05-15
add simple LFSR memory test after memory scan

XarkLabs:
Locally tested, seems useful.

### DIFF
--- a/code/software/memcheck/memcheck.c
+++ b/code/software/memcheck/memcheck.c
@@ -251,11 +251,9 @@ static KRESULT build_memory_map(volatile MEMINFO *header) {
 static void print_flags(uint32_t flags) {
   if ((flags & RAMBLOCK_FLAG_READONLY) != 0) {
     printf("[ROM] ");
-  }
-  else if ((flags & RAMBLOCK_FLAG_MAPPEDIO) != 0) {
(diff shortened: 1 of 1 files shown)


code/software/memcheck/memcheck.c
```diff
@@ -392,8 +421,137 @@ int main(int argc, char **argv) {
...
+        uint16_t v  = next_LFSR();
+        word_ptr[j] = v;
+      }
+    }
+
+    if (checkchar()) {
+      break;
+    }
+
+#if 0  // fake error testing
```
roscopeco, line 495: Just a question: do we want to keep this testing code?

XarkLabs: Well, maybe not, but I thought it was kind of useful for debugging to check error reporting "code path" so I left it in.

XarkLabs: If it concerns you I am happy to remove it. 🙂

roscopeco: No probs, was just curious in case it had been missed :) Happy for it to stay (and can see the use for it).

roscopeco (approved):
LGTM, thanks @XarkLabs 👍

PR #332, opened 2022-06-06
Add JLCPCB options to README

roscopeco:


0xTJ (approved):
Looks good, it's nice to have a set of reference options so that people can get their own PCBs made. :)

PR #336, opened 2022-07-02
Turn GCC up to 11

roscopeco:
Just updating GCC version in instructions to 11


XarkLabs (approved):
It has been working great for me. 👍

PR #337, opened 2022-07-08
Feature/extra duart support

roscopeco:
**Creating as a draft currently for RFC**


code/firmware/rosco_m68k_firmware/trap14.asm
```diff
@@ -99,7 +189,7 @@ FW_PRINT_C::
 ; It's cheaper size-wise to just duplicate this here...
 FW_PRINTLN_C::
     move.l  (4,A7),A0                 ; Get C char* from the stack into A0
-    bra.s   FW_PRINTLN                ; Call EARLY_PRINT
+    bra.s   FW_PRINTLN                ; Call FW_PRINT
```
0xTJ, line 192: One very minor thing here is that the comment doesn't match the code:
```suggestion
    bra.s   FW_PRINTLN                ; Call FW_PRINTLN
```

code/software/chardevs/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This is an empty (well, "Hello World") C project. You can use it as 
+a starting point for your own programs.
+ 
+## What next?
+
+* Edit `kmain.c` to add your own code.
+* Add more C or assembly files. Add them to the `$(OBJECTS)` line in `Makefile`
+  * Use the `.o` extension instead of `.c` or `.S` in the `Makefile`!
+* Make this README.md your own!
```
0xTJ, line 11: This `chardevs` example still has the original C Starter Project README and Makefile header comment. To avoid confusion, it'd be good to update the beginning part to match the code.

code/firmware/rosco_m68k_firmware/bootstrap.asm
```diff
@@ -211,6 +223,22 @@ EFP_DUMMY_NEGONE_D0L::
...
+    lsl.w   #2,D0
+    lea.l   DEVICE_BLOCKS,A0
+    bra.s   .START
+
+.LOOP:
+    clr.l   (A0)+
+
+.START:
+    dbra.w  D0,.LOOP
+    rts
```
0xTJ, line 239: Will this code clear the entire array? `DEVICE_BLOCKS` has a size of `C_DEVICE_SIZE * C_NUM_DEVICES = 0x20 * C_NUM_DEVICES`, but I think this will clear `((C_NUM_DEVICES << 2) + 1) * sizeof(long) = 0x10 * C_NUM_DEVICES + 4` bytes.

I haven't tested if this assembles (I know that some assemblers are picky about arithmetic) or works, but maybe something like this would work (though this requires a non-0 number of devices to clear):
```suggestion
INITDEVS:
    clr.w   DEVICE_COUNT    
    move.w  #C_DEVICE_SIZE*C_NUM_DEVICES-1,D0
    lea.l   DEVICE_BLOCKS,A0
    bra.s   .START

.LOOP:
    clr.b   (A0)+

.START:
    dbra.w  D0,.LOOP
    rts
```
Alternatively, adding a right-shift (or factoring out a multiple of 4) would allow the clear to still by long-wise, though it would require `C_DEVICE_SIZE` to be definitely divisible by 4.

code/firmware/rosco_m68k_firmware/InterfaceReference.md
```diff
@@ -44,6 +44,12 @@ as possible, and will be kept updated as firmware 2.0 is developed.
       * 1.2.2.5 PRINTCHAR (Function #4)
       * 1.2.2.6 SETCURSOR (Function #5)
       * 1.2.2.9 CHECKCHAR (Function #6)
+      * 1.2.2.8 CHECK_DEVICE_SUPPORT (Function #7)
+      * 1.2.2.9 GET_DEVICE_COUNT (Function #8)
+      * 1.2.2.10 GET_DEVICE (Function #9)
+      * 1.2.2.11 DEVICE_RECVCHAR (Function #10)
+      * 1.2.2.12 DEVICE_SENDCHAR (Function #11)
+      * 1.2.2.13 DEVICE_CHECKCHAR (Function #12)
```
0xTJ, line 52: It looks like there's a slight mismatch between this listing and the headings below:
```suggestion
      * 1.2.2.7 CHECKCHAR (Function #6)
      * 1.2.2.8 CHECK_DEVICE_SUPPORT (Function #7)
      * 1.2.2.9 GET_DEVICE_COUNT (Function #8)
      * 1.2.2.10 GET_DEVICE (Function #9)
      * 1.2.2.11 ADD_DEVICE (Function #10)
      * 1.2.2.12 DEVICE_RECVCHAR (Function #11)
      * 1.2.2.13 DEVICE_SENDCHAR (Function #12)
      * 1.2.2.14 DEVICE_CHECKCHAR (Function #13)
      * 1.2.2.15 RESERVED (Function #14)
      * 1.2.2.16 RESERVED (Function #15)
      * 1.2.2.17 DEVICE_CTRL (Function #16)
```

code/firmware/rosco_m68k_firmware/duart.asm
```diff
@@ -24,6 +24,9 @@
 ; If an MC68681 is found, this will set D5 to 1. Otherwise, D5 will be 0.
 ; This is used to signal INITSDB that it should use the DUART vectors instead of
 ; the MFP ones.
+;
+; Trashes: D0, A0-A5
+; Modifies: D5 (non-zero if DUART detected), MFP Regs
```
0xTJ, line 29: Minor comment thing: does this modify the MFP registers, or should it be this:
```suggestion
; Modifies: D5 (non-zero if DUART detected), DUART Regs
```

code/firmware/rosco_m68k_firmware/duart.asm
```diff
@@ -100,6 +121,34 @@ INITDUART::
     move.l  #EFP_DUMMY_NOOP,EFP_SETCURSOR
     move.l  #CHECKCHAR_DUART,EFP_CHECKCHAR
 
+    ; Setup device blocks...
+    lea.l   DEVICE_BLOCKS,A1
+    move.w  DEVICE_COUNT,D0
+    lsl.w   D0
```
0xTJ, line 127: Should this be:
```suggestion
    lsl.w   #5,D0
```
to account for the 0x20 size of the device blocks? I think this currently will write starting in the middle of an existing entry is any already exist.

code/software/libs/src/machine/machine.asm
```diff
@@ -1,172 +1,296 @@
...
+
+; Determine if character device support is in the firmware
+;
+; Trashes: Nothing
+; Modifies: D0 (return - 0 if unsupported, or any other value if supported)
+    section .text.mcCheckDeviceSupport
+mcCheckDeviceSupport::
+    move.l  D1,-(A7)
+    move.l  #7,D1
+    trap    #14
```
0xTJ, line 182: Although I really don't think it's likely to happen in practice, I think this will give a false positive if `D0` happens to already contain `0x1234FEDC`, since `TRAP #14` won't overwrite `D0` if the call is out of range of the table.
```suggestion
    moveq   #0,D0
    trap    #14
```

code/software/chardevs/README.md
roscopeco: Good spot, will do, thanks :)

code/firmware/rosco_m68k_firmware/bootstrap.asm
roscopeco: Hmm, yeah, I see what you mean, this will only clear half of the array (the left shift 2 is left over from before I increased the device size from 16 to 32). But I think just switching that to a left shift 3 will fix it - I don't see it "off by one" as I think you're suggesting (with the `-1`)?

I'll make the change now (and add a comments in the shared file about the size needing to be a multiple of 4) but happy to revisit if you still think it's wrong :)

code/firmware/rosco_m68k_firmware/duart.asm
roscopeco: Haha good spot, yes it should :D

roscopeco:
Rebased this and force pushed so if you have it checked out locally @0xTJ you might want to blow it away and re-pull.

code/firmware/rosco_m68k_firmware/bootstrap.asm
0xTJ: That looks good to me! :+1: Looking at it now, I have no idea why I thought the loop counter needed to be one less, I must have somehow ignored the branch to `.START` that was sitting right there. :sweat_smile:

0xTJ:
The changes look good! I'm going to give it another bit of review, and test things out, and then I'll hit the approve button. :smiley:

roscopeco:
Thanks for reviewing @0xTJ 👍

PR #339, opened 2022-08-02
Unify kernel file loading

0xTJ:
I noticed that the code for loading a kernel from mass storage was mostly duplicated between the IDE and BBSD, so this PR unifies that. It adds BBSD to the partition code, and passes partition handles for kernel loading.

Each valid partition on a device is checked in order, until a FAT partition containing `/ROSCODE1.BIN` is found, at which point the firmware tries to load it as before.

I'm not sure what consequences there are to handling partitions for SD, instead of the FAT library directly manipulating the whole device, but it does work with my SD cards. I also don't have the hardware to check that IDE still works as expected.

I'm happy to make significant changes to this PR, or not do this at all; this is more of an RFC.

This could also be done with a unified block device interface, instead of having the partition layer handle the difference device types. If that's added at some point in the future, this software loading would be changed to use that instead.

Doing this came out of taking a look at implementing ELF kernel loading.


roscopeco (approved):
Looks great @0xTJ , I like it. Thanks!

PR #341, opened 2022-09-10
Set GCC optimization parameter `min-pagesize=0`

### LINKED ISSUE #340: GCC gives false-positive array bounds error
On GCC 12.2.0, I get compile errors `-Werror=array-bounds`when dereferencing pointers created from literal constant addresses (like `(type *) 0x0000400`). This causes building the firmware to fail.

It seems that this is because the compiler treats addresses in the first page of memory as an indexed `NULL` pointer, which includes the SDB. This gives the diagnostics like the following:
```
main1.c: In function 'print_cpu_mem_info':
main1.c:120:12: error: array subscript 0 is outside array bounds of 'volatile SystemDataBlock[0]' [-Werror=array-bounds]
  120 |     if (sdb->cpu_model == 0) {
```
where `sdb` is defined as:
```
static volatile SystemDataBlock * const sdb = (volatile SystemDataBlock *)0x400;
```

0xTJ:
Using the GCC optimization option parameter to set `min-pagesize` to `0` fixes the problem in #340, seemingly by not having literal pointers in the first page be considered offset from `NULL`.

Fixes #340


XarkLabs (approved):
Nice find.

PR #342, opened 2022-09-10
Remove `-mno-strict-align` and `-mno-align-int`

0xTJ:
The option `-mno-strict-align` allows GCC to misalign values. In GCC 12.2, with `-mno-strict-align`, I've had issues appear where optimizations convert a "safe" unaligned word load (reading and shifting individual bytes) to a single word load, causing an address error.

`-mno-strict-align` causes GCC to assume that misaligned accesses are handled by the system.

`-mno-align-int` is also removed; that seems to be the default in the compiler anyways, since the opposite (`-malign-int`) is ABI-breaking.


XarkLabs (approved):
Nice, seems needed to me.

PR #343, opened 2022-09-12
Don't use custom build options for `stage2/load/load.o`

0xTJ:
When I unified loading the kernel in one spot, instead of duplicating code, I misunderstood the structure of the `include.mk` files, so `stage2/load/load.o` was being built without the right command-line options.

This led to GCC possibly generating MC68020+ opcodes, as encountered by SteveGG on Discord. The warning/error options were also not being included.

This PR leaves `stage2/load/load.o` to be built with the default pattern from the general `stage2` `Makefile`.

---

While this fixes the immediate issue, I would like to separately do a little cleanup in the firmware, both for any Makefiles that need it, and with the `extern` declarations, which could be put into headers.


XarkLabs (commented):
:+1:

XarkLabs (approved):
:+1: for real...

PR #345, opened 2022-10-08
Use unified Makefile include for software

0xTJ:
This PR moves the repeated code out of each software's `Makefile` into a unified `software.mk` that is included by all of them. The contents of `software.mk` is mostly the same as were each of the individual `Makefile`s, but all in one place.


0xTJ:
Will be updated after #347 is merged, to implement those fixes as the files are structured in this branch.

XarkLabs (approved):
moving forward!

PR #350, opened 2023-07-19
Fix `LOAD` segments having RWX permissions

0xTJ:
This PR fixes the `ld` warnings reported by Xark on Discord, such as:
```
warning: starter_c.elf has a LOAD segment with RWX permissions
```

This is caused by there being LOAD segments in the generated program that have both write and execute permissions, which could facilitate security exploits in programs. While these warnings are harmless here (we don't have an MMU yet to enforce permissions anyways), they are annoying. They can be suppressed with `--no-warn-rwx-segment` added to `LDFLAGS`, but this only hides the issue.

In this PR, I fix a couple few assembly files where the sections weren't being set to the right type. In vasm, the `data` is equivalent to `section .data,data`, instead of `section .data`. The former sets the section to have the properties of a data section, instead of a text one. This change prevents the `.data` and `.bss` from themselves becoming RWX.

The second cause of the issue here is that both text and data were being placed in the same pages in memory, which causes those pages to have permissions that are the union of those sections' permissions. To fix this, maximum-page-size-alignment is added between the text and data, to allow the linker to assign them different permissions.

Originally this wasted up to `0x1FFF` bytes of space in the binaries, but this was fixed by overriding the maximum page size to a smaller value. This will break loading in a system using an MMU, but saves considerable space. By default, this space saving is disabled for the '030, '040, and '060, which have version that include an MMU.

### DIFF
--- a/code/software/ehbasic/ehbasic.asm
+++ b/code/software/ehbasic/ehbasic.asm
@@ -9236,7 +9236,7 @@ LAB_SMSG
 * MID$	. MID$(<sexpr>,<nexpr>[,<nexpr>])				* done
 * USING$	. USING$(<sexpr>,<nexpr>[,<nexpr>]...])			* done
 
-	section .data
+	section .data,data
     align   12
 FREE_MEM:
 	
--- a/code/software/libs/src/start_serial/init.S
+++ b/code/software/libs/src/start_serial/init.S
@@ -130,7 +130,7 @@ CALL_DTORS::
     movem.l (A7)+,A2-A4
     rts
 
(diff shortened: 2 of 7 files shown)


Berzerkula (approved):
Reviewed, couldn't identify any malicious changes and all was described as in the Discord channel.

Berzerkula (approved):
I couldn't identify any malicious code changes and all looks changed according to the discussion in Discord channel.

XarkLabs (approved):
Approved!

Issue #307, opened 2022-02-02
HALT should be broken out to a pin somewhere


roscopeco:
As detailed in https://github.com/rosco-m68k/rosco_m68k/pull/306#issue-1117449775 , it is currently not possible to hard reset the system other than with the on-board button, as `/HALT` is not broken out to a pin anywhere. Raising this to fix that on a future board spin.

roscopeco:
PR #352

PR #352, opened 2023-08-17
Fix #333 (partial) and #307

### LINKED ISSUE #333: Proposed board alterations: Breakout EXPSEL in J6 and add optional DIP16 oscillator footprint
As requested, just jotting down two thoughts on potential improvements:

- At the moment EXPSEL is only broken out via J3, where adding it to J6 would make it possible to create expansion RAM hat PCBs that don't require IC2 to be customised where one doesn't want to use the J3 connector (for example, if you explicitly only want to use it for Xosera and don't want a backplane).
- For whatever reason DIP8 oscillators are getting rarer while DIP14 are still freely available. It looks like there's sufficient board space to add optional DIP14-compatible pins (similar to the way we accommodate either size of cap for the DUART crystal.

roscopeco:
* For #333
  * `EXPSEL` is now broken out onto a new pin on J6 (existing pins not moved for maximum chance of compatibility) 
  * `Q2` additional pins / footprint size / reroute is `WON'T DO`.
* For #307
  * `J8` is added with both `HALT` and `RESET`


roscopeco:
Screenshot 2023-08-18 at 00 21 23

0xTJ (approved):
Looks good to me! :+1: 
Happy to have those couple extra pins added.

PR #356, opened 2023-08-26
Switch to run on `pull_request` against `develop`.

roscopeco:


0xTJ (approved):
The change makes sense to me, and is still running the CI. :+1:

PR #357, opened 2023-08-26
updateflash from SD fix (ROM_BIN not defined)

XarkLabs:
Fix minor Makefile issue causing cp error when ROM_BIN not defined (e.g., update flash from ROM image on SD card, vs embedded in binary).


roscopeco (approved):
Thanks @XarkLabs !

PR #362, opened 2023-09-26
Improve Updateflash ROM ID Error Message

### LINKED ISSUE #361: Updateflash should mention JP3
The error message from updateflash correctly states when it cannot identify the ROM chips.

however, this is the case when the write protect jumper is open (a write is needed to put the chips in command mode). So if users have that jumper open and try to flash the chips, they’ll be told they have the wrong ROMs when they don’t.

the UX would be much improved if the error message mentioned the jumper 🙂

0xTJ:
This fixes #361 by adding a mention of the write protect jumper when the ROM size cannot be determined.

It also adds a check for the device manufacturer before checking the device ID, to avoid misidentifying a device with the same ID but a different manufacturer, which could be of a different size.


roscopeco (approved):
Looks good, thanks @0xTJ !

PR #365, opened 2023-10-09
Boot ELF Kernel from Storage

0xTJ:
I've had this mostly completed but just sitting around for a while. I hadn't properly tested it, and found that it didn't work with programs that copy themselves to 0x2000 (I had been testing with Mosys, which just runs from 0x40000). I've fixed that issue, so I'm looking for some thoughts on it.
This now works with the ELF files that are generated during the compilation of the software.

This loader only handles executable ELF files (not relocatable or dynamic objects). This is done to keep it simple (this is an alternative to just copying the flat binary to memory), only reading loadable segments to the physical memory addresses specified, not processing sections at all.

For now, the ELF segments are loaded to the physical addresses (LMA in the linker) given for each one, and (the default) 0x40000 is kept as the address to run. Ideally, every segment would be loaded to its virtual address, and execution would start at the address given in E_ENTRY. Since stage2 is running from 0x2000, it conflicts with the virtual address (VMA in the linker) of typical programs. For now, the program init is responsible for moving the rest of the program to its run location, as with a normal flat binary program.
This is controlled by the ELF_LOAD_VIRTUAL_ADDR macro.

The linker script entry point for software is changed to be the first bit of code that doesn't copy the program to 0x2000. Currently this has no effect, but it will be meaningful if the program segments are loaded to the correct virtual addresses, in which case E_ENTRY will be used.

This requires using more than the minimal FATFS API, but this change by itself doesn't significantly increase code size.
For this PR, I'm looking for some comments on the approach. I'm not fully satisfied with it, but I think it may be harder to get where I'd like easily while stage2 gets decompressed to RAM, instead of running from flash.

### DIFF
--- a/code/firmware/rosco_m68k_firmware/stage2/load/include/elf.h
+++ b/code/firmware/rosco_m68k_firmware/stage2/load/include/elf.h
@@ -1,0 +1,123 @@
+/*
+ *------------------------------------------------------------
+ *                                  ___ ___ _
+ *  ___ ___ ___ ___ ___       _____|  _| . | |_
+ * |  _| . |_ -|  _| . |     |     | . | . | '_|
+ * |_| |___|___|___|___|_____|_|_|_|___|___|_,_|
+ *                     |_____|       firmware v2
+ * ------------------------------------------------------------
+ * Copyright (c)2020-2022 Thomas Jager
+ * See top-level LICENSE.md for licence information.
+ *
+ * Definitions for ELF file contents.
+ * Only contains the ELF Header and Program Header.
+ * 
+ * Structures, names, and values based on:
+ *   Tool Interface Standard (TIS)
+ *   Executable and Linking Format (ELF) Specification
+ *   Version 1.2
+ * ------------------------------------------------------------
+ */
+
+#ifndef __ROSCO_M68K_ELF_H
+#define __ROSCO_M68K_ELF_H
+
+#include <stdint.h>
+
+#pragma pack(push, 1)
+
+typedef uint32_t Elf32_Addr;
+typedef uint16_t Elf32_Half;
+typedef uint32_t Elf32_Off;
+typedef int32_t Elf32_Sword;
+typedef uint32_t Elf32_Word;
+
+// ELF Header
+
+#define EI_MAG0         0
+#define EI_MAG1         1
+#define EI_MAG2         2
+#define EI_MAG3         3
+#define EI_CLASS        4
+#define EI_DATA         5
+#define EI_VERSION      6
+#define EI_PAD          7
+#define EI_NIDENT       16
+
+#define ELFMAG0     0x7f
+#define ELFMAG1     'E'
+#define ELFMAG2     'L'
+#define ELFMAG3     'F'
+
+#define ELFCLASSNONE    0
+#define ELFCLASS32      1
+#define ELFCLASS64      2
+
+#define ELFDATANONE 0
+#define ELFDATA2LSB 1
+#define ELFDATA2MSB 2
+
+#define ET_NONE     0
+#define ET_REL      1
+#define ET_EXEC     2
+#define ET_DYN      3
+#define ET_CORE     4
+#define ET_LOPROC   0xff00
+#define ET_HIPROC   0xffff
+
+#define EM_NONE         0
+#define EM_M32          1
+#define EM_SPARC        2
+#define EM_386          3
+#define EM_68K          4
+#define EM_88K          5
+#define EM_860          7
+#define EM_MIPS         8
+#define EM_MIPS_RS4_BE  10
+
+#define EV_NONE     0
+#define EV_CURRENT  1
+
+typedef struct {
+    unsigned char   e_ident[EI_NIDENT];
+    Elf32_Half      e_type;
+    Elf32_Half      e_machine;
+    Elf32_Word      e_version;
+    Elf32_Addr      e_entry;
+    Elf32_Off       e_phoff;
+    Elf32_Off       e_shoff;
+    Elf32_Word      e_flags;
+    Elf32_Half      e_ehsize;
+    Elf32_Half      e_phentsize;
+    Elf32_Half      e_phnum;
+    Elf32_Half      e_shentsize;
+    Elf32_Half      e_shnum;
+    Elf32_Half      e_shstrndx;
+} Elf32_Ehdr;
+
+// Program Header
+
+#define PT_NULL     0
+#define PT_LOAD     1
+#define PT_DYNAMIC  2
+#define PT_INTERP   3
+#define PT_NOTE     4
+#define PT_SHLIB    5
+#define PT_PHDR     6
+#define PT_LOPROC   0x70000000
+#define PT_HIPROC   0x7fffffff
+
+typedef struct {
+    Elf32_Word  p_type;
+    Elf32_Off   p_offset;
+    Elf32_Addr  p_vaddr;
+    Elf32_Addr  p_paddr;
+    Elf32_Word  p_filesz;
+    Elf32_Word  p_memsz;
+    Elf32_Word  p_flags;
+    Elf32_Word  p_align;
+} Elf32_Phdr;
+
+#pragma pack(pop)
+
+#endif
--- a/code/firmware/rosco_m68k_firmware/stage2/load/include/elf.h
+++ b/code/firmware/rosco_m68k_firmware/stage2/load/include/elf.h
@@ -24,8 +24,6 @@
 
 #include <stdint.h>
 
-#pragma pack(push, 1)
-
 typedef uint32_t Elf32_Addr;
 typedef uint16_t Elf32_Half;
 typedef uint32_t Elf32_Off;
@@ -93,7 +91,7 @@ typedef struct {
     Elf32_Half      e_shentsize;
     Elf32_Half      e_shnum;
     Elf32_Half      e_shstrndx;
-} Elf32_Ehdr;
+} __attribute__ ((packed)) Elf32_Ehdr;
 
 // Program Header
 
@@ -116,8 +114,6 @@ typedef struct {
     Elf32_Word  p_memsz;
     Elf32_Word  p_flags;
     Elf32_Word  p_align;
-} Elf32_Phdr;
-
-#pragma pack(pop)
+} __attribute__ ((packed)) Elf32_Phdr;
 
 #endif
--- a/code/firmware/rosco_m68k_firmware/stage2/load/load.c
+++ b/code/firmware/rosco_m68k_firmware/stage2/load/load.c
@@ -21,6 +21,14 @@
 #include "part.h"
 #include "system.h"
 
+/*
+ * Currently, software that relocates itself to 0x2000 must be loaded like a
+ * raw binary program (to 0x40000), to avoid overwriting firmware stage2.
+ * If ELF_LOAD_VIRTUAL_ADDR is 1, instead load every segment to its run virtual
+ * address, and use E_ENTRY.
+ */ 
+#define ELF_LOAD_VIRTUAL_ADDR 0
+
 extern void mcPrint(const char *str);
 extern void print_unsigned(uint32_t num, uint8_t base);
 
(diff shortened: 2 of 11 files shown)


code/firmware/rosco_m68k_firmware/stage2/load/include/elf.h
```diff
@@ -0,0 +1,123 @@
...
+ *   Version 1.2
+ * ------------------------------------------------------------
+ */
+
+#ifndef __ROSCO_M68K_ELF_H
+#define __ROSCO_M68K_ELF_H
+
+#include <stdint.h>
+
+#pragma pack(push, 1)
```
roscopeco, line 27: Minor; _Generally_ I prefer `__attribute__((packed))` on the individual structures, as that's where I tend to look (less cognitive load than having to scroll around and see what's been pushed / popped).

code/firmware/rosco_m68k_firmware/stage2/load/load.c
```diff
@@ -15,17 +15,39 @@
```
roscopeco, line 15: So, looking at this, I _think_ the elf loading _replaces_ the flat binary loader, is that right?

code/firmware/rosco_m68k_firmware/stage2/load/load.c
```diff
@@ -15,17 +15,39 @@
...
+ * address, and use E_ENTRY.
+ */ 
+#ifndef ELF_LOAD_VIRTUAL_ADDR
+#define ELF_LOAD_VIRTUAL_ADDR 0
+#endif
+
 extern void mcPrint(const char *str);
 extern void print_unsigned(uint32_t num, uint8_t base);
 
+typedef void (*KMain)(volatile SystemDataBlock * const);
```
roscopeco, line 37: Is this now in two places (this file, and the one below)? If so, just wondering if it's worth moving it to a header or something?

roscopeco (commented):
I suspect I'm probably missing something important here (so please forgive me if my comments don't make sense 😂 ). I feel like this is _replacing_ the flat binary loader, but then looking at the link-script changes and the moving of the entry point, I get the impression that it's actually still supporting the flat loader (which will still do the usual relocation) _as well_ as the elf loader, which... won't get relocated to low mem?

code/firmware/rosco_m68k_firmware/stage2/load/load.c
0xTJ: It doesn't fully replace the flat binary loading, it just adds another option. At line 315, it adds a second file to check for and try to load. It will still first try to load `ROSCODE1.BIN` as a flat binary, but will then try to read `ROSCODE1.ELF` an ELF file.

roscopeco: I see, that makes more sense to me now, thanks! 👍

0xTJ:
Thanks for the initial review! I didn't want to make it fully incompatible with the old approach, so I avoided completely replacing the flat binary loader, just adding a second file option (`ROSCODE1.ELF`).

Ideally, it would be able to just load arbitrary ELF files right to where they need to run at, but stage2 is decompressed to that location (0x2000). I've tried to set it up so that switching to directly loading to the run address is easy, but that would require that either stage2 be located elsewhere (some other bit of reserved RAM, or uncompressed in ROM), or segments that overlap with stage2 are kept elsewhere, then copied in during the switch from stage2 to the program.

I'm not particularly happy that this then gives effectively no benefit right now over just loading a flat binary (when `ELF_LOAD_VIRTUAL_ADDR` is 0), so if you have any ideas on how to handle this I'd appreciate them.

roscopeco:
> I suspect I'm probably missing something important here [...]

So, I _think_ I get it now, based on your answer to another comment, and my having read the description more closely. Just to check my understanding;

* Flat binary loading remains the same; If a .bin file is sent, it will be loaded in the usual way, including relocating itself to `0x2000` (or whatever, depending on the init that it's linked with).
* ELF loading is new - if the file sent is a `.elf` it's loaded at the physical address specified in the ELF (which for the entry point **must** be `0x40000` for compatibility, but data segments etc can be anywhere as long as they're not stomping system memory).
* There's an option to load at the VMA, disabled currently. So elf binaries must be linked to run at their PMA (and we don't do relocation etc)?

If the above is correct, I am still slightly confused about the changes in the link scripts:

* I'm assuming those are still going to link flat binary (since they still specify that output format)?
* If so, I'm not understanding why we've changed the entry point to skip the copying to low mem?
  * (Or perhaps we haven't and I still don't quite get it 😅)

roscopeco:
> Thanks for the initial review! I didn't want to make it fully incompatible with the old approach, so I avoided completely replacing the flat binary loader, just adding a second file option (`ROSCODE1.ELF`).
> 
> Ideally, it would be able to just load arbitrary ELF files right to where they need to run at, but stage2 is decompressed to that location (0x2000). I've tried to set it up so that switching to directly loading to the run address is easy, but that would require that either stage2 be located elsewhere (some other bit of reserved RAM, or uncompressed in ROM), or segments that overlap with stage2 are kept elsewhere, then copied in during the switch from stage2 to the program.
> 
> I'm not particularly happy that this then gives effectively no benefit right now over just loading a flat binary (when `ELF_LOAD_VIRTUAL_ADDR` is 0), so if you have any ideas on how to handle this I'd appreciate them.

Right, gotcha, thanks 👍 . My thoughts (somewhat off the top of my head, with one eye on the future):

* I think even with the limited loading just at physical for now, this still adds value. Building elf is one less step / simpler makefiles, and it gives us scope to expand this in the future. Ultimately I'd like us to get away from fixed load addresses entirely, and I think without an MMU VMA will probably never make any more sense the PMA does (it's just swapping one physical area for another). However, there's scope here to build in relocation to arbitrary (physical) addresses at some later time - which I'm thinking is where I'd like us to get to eventually.
* In terms of moving stage2:
    * I'd have no issue with having it uncompressed in ROM _in principle_ - we certainly have the ROM space to burn now, which was the driver for compressing it in the first place - if it weren't for the fact that I'm still keen to provide backward compatibility for older boards. That said, there's probably a relatively small user-base for that (maybe we should do a poll of how many people are running old boards with current firmware 🤔 ) so it _might_ be worth investigating. 
    * I also wouldn't be against decompressing it to somewhere more convenient (and that should have zero impact on anyone except us in real terms). The question would then become "where would it be better"? Reserve a few KiB at the top of physical RAM, and move the stack below it maybe?
* Obviously there'd still be _some_ areas off limits (the very bottom of memory where the vector table and firmware structures / bus live - some of which needs to remain where it is for backwards compatibility too).

0xTJ:
> There's an option to load at the VMA, disabled currently. So elf binaries must be linked to run at their PMA (and we don't do relocation etc)?

It's a bit confusing because of the way it reuses terms, but the virtual and physical addresses in the ELF file in this case don't *exactly* correspond to virtual and physical addresses. The ELF virtual addresses have values from the VMA (run addresses) in the linker scripts. The ELF physical addresses, instead of being actual physical addresses, are the LMA (load addresses) from the linker script (starting from 0x40000).
The binaries are linked to run based at the same address they were before (VMA starting at 0x2000).

So the LMA is the space where the flat binary loader would copy the program, while the VMA is the space you're in by the time you've hit `POSTINIT` in `init.S`. (Hopefully the paragraph above hasn't just made things more confusing :sweat_smile:)

> If so, I'm not understanding why we've changed the entry point to skip the copying to low mem?

That change is because the entry point in the linker is a VMA address, which is relevant after the code has been copied (whether by `init.S` or by an ELF loader). Previously, the entry point was to the code that copies the program from 0x40000 to 0x2000, but the value of E_ENTRY was actually 0x2000, since that's where the code is linked to run at.

If you use the init code that copies from the load location to the run location as the entry point for the ELF loader, it'll (at best) copy the data to the location where it already exists. 0x40000 is still the point at which execution starts for both the flat binary loader, and for the case where `ELF_LOAD_VIRTUAL_ADDR` is 0 (as it is now), but if `ELF_LOAD_VIRTUAL_ADDR` is 1 the code that copies from 0x40000 to 0x2000 should be skipped.

0xTJ:
>       * I also wouldn't be against decompressing it to somewhere more convenient (and that should have zero impact on anyone except us in real terms). The question would then become "where would it be better"? Reserve a few KiB at the top of physical RAM, and move the stack below it maybe?

At least as a first step, this would be my preferred approach for keeping stage2 out of the way. I think that it's cleaner to reuse that memory, compared to starting at 0x2000, since the software just needs to reset the stack pointer as it already does.

If we take this approach, then I could clean up the code to always (for ELF) load directly to the run address, instead of to 0x40000, skipping the extra copy step and allowing for much more flexibility.

roscopeco:
> > There's an option to load at the VMA, disabled currently. So elf binaries must be linked to run at their PMA (and we don't do relocation etc)?
> 
> It's a bit confusing because of the way it reuses terms, [...] The binaries are linked to run based at the same address they were before (VMA starting at 0x2000).
>
> So the LMA is the space where the flat binary loader would copy the program, while the VMA is the space you're in by the time you've hit `POSTINIT` in `init.S`. (Hopefully the paragraph above hasn't just made things more confusing 😅)
> 

Right, apologies, I'm adding to that confusion. I was thinking of the same things you describe, but my head is in "kernel land" at the moment so I used those terms instead 😅. I think we're on the same page though. 

> > If so, I'm not understanding why we've changed the entry point to skip the copying to low mem?
> 
> That change is because the entry point in the linker is a VMA address, which is relevant after the code has been copied (whether by `init.S` or by an ELF loader). Previously, the entry point was to the code that copies the program from 0x40000 to 0x2000, but the value of E_ENTRY was actually 0x2000, since that's where the code is linked to run at.
>

Cool, I think I get that now. We're changing that entry point:

* so if someone switches the output format to elf, the entry point will be correct, and
* it doesn't matter for flat-binary, which doesn't particularly care about that entry point anyway (and we manually link the init section at the start of the binary anyway).

Is that correct?
 
> If you use the init code that copies from the load location to the run location as the entry point for the ELF loader, it'll (at best) copy the data to the location where it already exists. 0x40000 is still the point at which execution starts for both the flat binary loader, and for the case where `ELF_LOAD_VIRTUAL_ADDR` is 0 (as it is now), but if `ELF_LOAD_VIRTUAL_ADDR` is 1 the code that copies from 0x40000 to 0x2000 should be skipped.

Gotcha 👍  (with caveat above that I might still be understanding wrong 😅 ).

roscopeco:
> > ```
> >   * I also wouldn't be against decompressing it to somewhere more convenient (and that should have zero impact on anyone except us in real terms). The question would then become "where would it be better"? Reserve a few KiB at the top of physical RAM, and move the stack below it maybe?
> > ```
> 
> 
> At least as a first step, this would be my preferred approach for keeping stage2 out of the way. I think that it's cleaner to reuse that memory, compared to starting at 0x2000, since the software just needs to reset the stack pointer as it already does.

Cool, okay, then that option gets my vote. I can't think of any reason anyone outside of firmware devs would care, but I'll pop a question (and link to this conversation) in Discord in case anyone does.

> If we take this approach, then I could clean up the code to always (for ELF) load directly to the run address, instead of to 0x40000, skipping the extra copy step and allowing for much more flexibility, with the flat binary loading unaffected.

Sounds reasonable, again, it would get my vote.

0xTJ:
> Cool, I think I get that now. We're changing that entry point:
> 
>     * so if someone switches the output format to elf, the entry point will be correct, and
> 
>     * it doesn't matter for flat-binary, which doesn't particularly care about that entry point anyway (and we manually link the init section at the start of the binary anyway).
> 
> 
> Is that correct?

Yeah, exactly! :+1: 

> > If you use the init code that copies from the load location to the run location as the entry point for the ELF loader, it'll (at best) copy the data to the location where it already exists. 0x40000 is still the point at which execution starts for both the flat binary loader, and for the case where `ELF_LOAD_VIRTUAL_ADDR` is 0 (as it is now), but if `ELF_LOAD_VIRTUAL_ADDR` is 1 the code that copies from 0x40000 to 0x2000 should be skipped.
> 
> Gotcha 👍 (with caveat above that I might still be understanding wrong 😅 ).

It's definitely a bit confusing because we end up talking about the flat binary loader, the "pseudo-flat" ELF loader (`ELF_LOAD_VIRTUAL_ADDR` = 0), and the full ELF loader that copies everything where it needs to be (`ELF_LOAD_VIRTUAL_ADDR` = 1). Hopefully no issues come up with moving stage2, and this messy `ELF_LOAD_VIRTUAL_ADDR` becomes irrelevant anyways.

roscopeco (approved):
I'm gonna go with 24 hours being a long enough consultation period on this, let's do it!

Thanks again for this @0xTJ , it's awesome 🚀 💪

roscopeco (approved):
Looks great, thanks! 🥳

PR #366, opened 2023-10-09
Move Decompressed Stage2 to the Top of On-Board RAM

0xTJ:
Pending any issues that are raised with doing it, this PR moves the decompressed stage2 firmware from 0x2000 to 0xF0000. The stack pointer (in firmware) is adjusted to start below that point instead of at the top of on-board memory.

Software `start_serial` is unchanged, and will still reset the stack pointer to the value at `SDB_MEMSIZE`.

Since kermit requires large buffers, these are placed starting at 0x2000. Since kermit-received files always go to 0x40000, this keeps the buffers out of the way, but allows this memory to be used for other purposes when kermit isn't active (e.g. ELF segments loaded there).

### DIFF
--- a/code/firmware/rosco_m68k_firmware/bootstrap.asm
+++ b/code/firmware/rosco_m68k_firmware/bootstrap.asm
@@ -20,7 +20,7 @@
     section .text
 
 VECTORS:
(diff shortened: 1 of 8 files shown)


roscopeco (approved):
Looks great! Thanks! 🥳

PR #372, opened 2023-10-24
ROM kernel 101

roscopeco:
Initial integration of the ROM kernel.

Apols for the **enormous** PR, but this stuff has to come from somewhere 🤣


code/firmware/rosco_m68k_firmware/kernel/tests/Makefile
0xTJ, line 1: This file causes build errors for me because it sets and uses `GCC_LIBS` without allowing it to be set by the top-level `user.mk`.

With the compiler and libraries installed through AUR on Arch, `GCC_LIBS` needs to not have `m68000/` appended to each entry. Making the following change (matching `software.mk`) allows it to build:
```diff
@@ -1,4 +1,6 @@
 ROSCO_M68K_DIR=../../../../..
+-include $(ROSCO_M68K_DIR)/user.mk
+
 SRC_DIR=..
 R68K_DIR=../../../../tools/r68k
 SYSLIBROOT=$(ROSCO_M68K_DIR)/code/software/libs
@@ -31,7 +33,7 @@ ASFLAGS=-mcpu=$(CPU) -march=$(CPU)
 VASMFLAGS=-Felf -m$(CPU) -quiet -Lnf $(DEFINES)
 ARFLAGS=
 LIBS=-lrosco_m68k -lgcc
-GCC_LIBS=$(shell $(CC) --print-search-dirs 											\
+GCC_LIBS?=$(shell $(CC) --print-search-dirs 										\
 		| grep libraries:\ =														\
     | sed 's/libraries: =/-L/g' 													\
     | sed 's/:/m68000\/ -L/g')m68000/
```

code/firmware/rosco_m68k_firmware/main1.c
```diff
@@ -114,9 +122,27 @@ static void initialize_loader_efp() {
     *program_loader_ptr = (uint32_t)default_program_loader;
 }
 
+#ifdef WITH_KERNEL
+static void initialize_kernel(void) {
+    api_init();
+    pmm_init();
+    pmm_free(0x20000, 0x30000);
```
0xTJ, line 129: Will giving this range to the kernel cause issues with software running? When programs copy themselves from 0x40000 down to 0x2000, they could overlap with this region.

0xTJ (commented):
I haven't had a chance to go through all of it yet, but these are a couple initial comments.

code/firmware/rosco_m68k_firmware/kernel/tests/Makefile
roscopeco: Cool, will do, thanks :) 👍

code/firmware/rosco_m68k_firmware/main1.c
roscopeco: After further reflection ([I'd forgotten](https://discord.com/channels/698525682199822367/711035186170822727/1166979412856668241) about things that work the way `sdfat_menu` does) I've moved this responsibility into the program for the time being.

0xTJ: Sounds good :+1: It is annoying to have to do it this way (and it not "just work"), but I think it would have to be a (theoretically) breaking change for `BIN`-loaded programs, with everything beyond 0x2000 being owned by the software.

code/tools/r68k/main.cpp
```diff
@@ -0,0 +1,543 @@
...
+            buf[bp--] = digit(num % base);
+            num /= base;
+        }
+    } 
+
+    return ((char*)&buf[bp+1]);
+}
+
+extern "C" {
+    rosco::m68k::emu::AddressDecoder* __mem;
```
0xTJ, line 99: For a program that's being built for the host, I think that it'd be better to avoid using implementation-reserved identifiers.

0xTJ:
I've finished reviewing, and it looks good! I haven't noticed any issues, and it seems to be working fine for me!
I added one more comment, but once that's resolved (with or without a change), I'm good to approve if nothing else comes up! :+1:

code/tools/r68k/main.cpp
roscopeco: Good shout, fixed 👍

0xTJ (approved):
As far as I can tell, it looks good, approving! :+1:

PR #377, opened 2023-10-28
Cleanup Outdated Equates

0xTJ:
I noticed that there were some outdated equates in `equates.S` and in `rosco_m68k_public.asm`, having values still set for rev. 1 boards.

`equates.S` is only used in software in `markm-ide/hdfat_menu`, and I don't think the values are used there are all.
In firmware, it's only used in the `attic` versions, which don't point to it anyways because they've been moved one directory deeper. One option could be to add a `shared/equates.S` in the `attic` directory. It still wouldn't build because the location is wrong, but it would keep the contents around for anyone in the future. The alternative would be to do nothing, and let those break, requiring the use of git tags to go back and see the contents of them in the future. I'm looking for opinions on how to handle that.

`rosco_m68k_public.asm` also had some rev.-1-specific equates in it. Since the only place that `IOBASE` is ever used in the repo is in `rosco_m68k_public.asm` itself (for `MFPBASE`), I've just replaced the `MFPBASE` `equ` with the value directly, and removed the memory-layout-dependent defines from that file (to keep the public interface as generic as possible). Only the `RAMBASE` and `RAMLIMIT` were moved into the private interface (though only `RAMLIMIT` is ever used), since the others are revision-specific and not used. ~This change also breaks the spirit of the `attic` firmware versions, since some of the older ones use `ROMBASE` (though like with `equates.S`, they don't actually point to this file, so they can't build anyways).~


roscopeco:
Cool, thanks @0xTJ , nice bit of cleanup 👍 

I'm wondering if we should just go ahead and delete the `attic` from the firmware directory now... It's in git history and the release branches are still there... Seems like deadweight...

roscopeco (approved):
LGTM 👍

0xTJ:
Fully agree on deleting `attic`, it's all there anyways, might as well not drag extra copies around. :+1:

PR #378, opened 2023-10-28
Fix #359

### LINKED ISSUE #359: RN1 value in 68020 BOM should be 4k7
10k is probably fine (and will save on BOM cost) so it’s probably worth doing some actual tests (it only really affects interrupt lines on the expansion connector, plus cache inhibit and a few other CPU lines have a stronger pull-up). 

However the actual designed value is 4k7 so should either test and rerate to 10k in the design, or fix the BOM to be 4k7.

roscopeco:
Change out RN1 value in 020 BOM to 4k7 to match schematic


0xTJ:
Can you confirm that `docs/r2_020_r1.1_bom.numbers` does have the BOM in it? When I open it in LibreOffice, I see the BOM part being blank, with  just the footer showing. The other Numbers spreadsheets in that directly have contents, just not that one.

roscopeco:
> Can you confirm that `docs/r2_020_r1.1_bom.numbers` does have the BOM in it? When I open it in LibreOffice, I see the BOM part being blank, with just the footer showing. The other Numbers spreadsheets in that directory have contents, just not that one.

Hmm, weird. Seems good here 🤔 

Screenshot 2023-10-30 at 20 46 03

0xTJ (approved):

PR #379, opened 2023-10-28
Delete firmware attic

roscopeco:


0xTJ (approved):

0xTJ:
Happy to have that cleaned up! :+1:

PR #380, opened 2023-10-28
Add CRC7 and 16 to SD Card routines

harrowm:
This one needs some testing by someone on different kit/SD Cards to mine.  I have tested on a 020 and 010 board with FW 2.4 and a cheap 8Gb and Samsung 64Gb SD Card.

Maybe I should have feature toggled this via a #define .. let me know if I add that in.

My SD Card set up is "better" but not perfect.  I don't think this code makes the SD Card 100% reliable on my setup, but I also don't think it does any harm to have it there ..


roscopeco (changes requested):
Hmm, this seems to have introduced issues for me

Tested here with 2 SDHC cards (both 32GB, one Kingston, one Phillips) and one SDv2 (2GB, transcend). All working reliably across multiple reads and writes (using sdfat_menu) under development firmware.

With this change, the only card that came close to working was the Kingston, which did at least write and verify some of the file (though never the whole thing). I noticed writes were _significantly_ slower here too.

The other two cards "wrote" very (suspiciously) quickly and then consistently failed immediately at sector zero on the read back.

Screenshot 2023-11-11 at 19 02 16

harrowm:
Omg.  Let me recheck :oThanks for spending the time on this. On Nov 11, 2023, at 2:05 PM, Ross Bamford ***@***.***> wrote:﻿
@roscopeco requested changes on this pull request.

Hmm, this seems to have introduced issues for me
Tested here with 2 SDHC cards (both 32GB, one Kingston, one Phillips) and one SDv2 (2GB, transcend). All working reliably across multiple reads and writes (using sdfat_menu) under development firmware.
With this change, the only card that came close to working was the Kingston, which did at least write and verify some of the file (though never the whole thing). I noticed writes were significantly slower here too.
The other two cards "wrote" very (suspiciously) quickly and then consistently failed immediately at sector zero on the read back.

code/firmware/rosco_m68k_firmware/blockdev/bbsd.c
```diff
@@ -197,12 +287,13 @@ bool BBSD_read_block(BBSDCard *sd, uint32_t block, uint8_t *buffer) {
...
     BBSPI_recv_buffer(buffer, 512);
+    int checksum = ((BBSPI_recv_byte() << 8) | BBSPI_recv_byte()) & 0xFFFF;
 
-    // Get checksum too (and ignore it ;) )
-    BBSPI_recv_byte();
-    BBSPI_recv_byte();
-
-    result = true;
+    // Check the checksum.  Checksum stored big endian on sd card 
+    // Allow for any legacy rosco SD Cards written with a dummy checksum
```
0xTJ, line 293: I don't think this comment on legacy rosco_m68k cards having been written with a dummy checksum is correct. As far as I know, the CRC is used to protect against over-the-wire errors, not the integrity of blocks on disk, so I don't think that `0xFFFF` being read is related to `0xFFFF` having been written.

harrowm:
You think the card is not storing the crc16 but recalculating and comparing to what’s been sent ?  Do all cards support crc16 ? Is it optional ?  (Or if they don’t support they just ignore?) On Nov 12, 2023, at 8:35 AM, Thomas Jager ***@***.***> wrote:﻿
@0xTJ commented on this pull request.



In code/firmware/rosco_m68k_firmware/blockdev/bbsd.c:
>  
-    // Get checksum too (and ignore it ;) )
-    BBSPI_recv_byte();
-    BBSPI_recv_byte();
-
-    result = true;
+    // Check the checksum.  Checksum stored big endian on sd card 
+    // Allow for any legacy rosco SD Cards written with a dummy checksum

I don't think this comment on legacy rosco_m68k cards having been written with a dummy checksum is correct. As far as I know, the CRC is used to protect against over-the-wire errors, not the integrity of blocks on disk, so I don't think that 0xFFFF being read is related to 0xFFFF having been written.

code/firmware/rosco_m68k_firmware/blockdev/bbsd.c
0xTJ: > You think the card is not storing the crc16 but recalculating and comparing to what’s been sent ?  Do all cards support crc16 ? Is it optional ?  (Or if they don’t support they just ignore?) 

Yes, from the SD Physical Layer Simplified Specification, the CRC used to protect against transmission errors, and it's generated by the card. By default, CRC should be disabled when the card enters SPI mode (with some exceptions, like CMD8), unless it's enabled by using CMD59. With this PR removing the CMD59 that was present, I believe that compliant cards *should* be in CRC OFF mode, but the comment that was there implies that some cards were defaulting to staying in CRC ON without it.

I think that CRC support is a basic feature of SD, and should be present on all cards (it's always enabled in SDIO mode, so the CMD0 to initially switch to SPI must have a valid CRC). The CRC_ON_OFF command (CMD59) is described as a mandatory command to be supported, and is in the "basic" command class.

roscopeco:
Are we still planning to keep this one @harrowm @0xTJ or should we close it?

harrowm:
We fixed my issue by changing the timeout.  We can close this.  Is this something I should do ?Sent from my iPhoneOn Sep 6, 2024, at 3:33 PM, Ross Bamford ***@***.***> wrote:﻿
Are we still planning to keep this one @harrowm @0xTJ or should we close it?

roscopeco:
> We fixed my issue by changing the timeout.  We can close this.  Is this something I should do ?Sent from my iPhoneOn Sep 6, 2024, at 3:33 PM, Ross Bamford ***@***.***> wrote:﻿ Are we still planning to keep this one @harrowm @0xTJ or should we close it? —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

Great stuff, glad to hear it got fixed @harrowm 🙂

PR #382, opened 2023-11-07
Speed up write performance

harrowm:
While using the r68k tool to test writing files, I noticed that the write performance of writing a new file is very bad.  On a 64Gb image, 'write test" takes about 3 seconds on my machine (a M1 Air) if the file test exists, but about 170 seconds if it does not.

By logging the reads/write in r68k, it's possible to see that if the file exists r68k has to do 1,086 sector reads and 1,024 sector writes but if the file does not exist then r68k does 296,134 reads and 9,227 writes.

The cause of this is that the fat library we use doesn't use the FS_HINTS sector when allocating new entry in the fat table.  This PR changes the fat_io library to use the "next free cluster" hint.

Doing this reduces the reads/writes to write a new test file to 4,114 and 11,662.  This speeds up execution from 172 seconds before the change to about 3.75 seconds.

I have only tested this change under r68k, not a real rosco board.


roscopeco (approved):
Looks good, and seems okay on hardware. Thanks 👍

PR #384, opened 2023-11-11
Fix outdated IC14 reference

roscopeco:
Reported in Discord by kittymeow : https://discord.com/channels/698525682199822367/698527817259941970/1172907116365885450


0xTJ:
One alternative (at least in KiCad 7, I think it was also there in 6) to directly writing `IC5` in the text box is writing `${IC5:REFERENCE}`. This gets stored as a reference to that particular symbol instance, so if the reference designator gets changed, the text will be updated to match the new name (if you delete that symbol, you end up with `${<something like a UUID>:REFERENCE}`. You can get a number of properties of symbols doing this, but I've used the `REFERENCE` the most in text comments.

roscopeco:
> One alternative (at least in KiCad 7, I think it was also there in 6) to directly writing `IC5` in the text box is writing `${IC5:REFERENCE}`. This gets stored as a reference to that particular symbol instance, so if the reference designator gets changed, the text will be updated to match the new name (if you delete that symbol, you end up with `${<something like a UUID>:REFERENCE}`. You can get a number of properties of symbols doing this, but I've used the `REFERENCE` the most in text comments.

Good shout @0xTJ , thanks 🙂 ! Done 👍

0xTJ (approved):
Looks good, approving! :+1:

PR #387, opened 2023-11-19
Feature/68000 probe support

roscopeco:
Happy to discuss this one, it's kind of a big change (conceptually, if not actually from the code POV).

* Support device probing on 68000 CPUs (with a bit of stack hackery in the bus error handler)
* Switch default CPU for Firmware build to be "lowest comment denominator" 68000
* Switch default CPU for software to 68000 for the same reasons
* Fail fast (and polite) in memcheck on 68000

Additionally, this:

* Prepares ATA / IDE for (possible) future deprecation / removal from standard build

This just means I made a new make option `WITH_ATA` - I needed this bit while developing, and it seemed like it could be useful down the line so left it in. It's on by default (so no change).

I have tested this on 68000, 68010 and it seems fine. ~If anyone has a populated 020 handy it would be good to double check it hasn't broken anything there~ **Edit: This has now been successfully tested on 020 as well**. 

Proof of life, no rebuild between these runs - just yanking and moving the chips:

Screenshot 2023-11-19 at 21 47 36

### DIFF
--- a/code/firmware/rosco_m68k_firmware/blockdev/ata.c
+++ b/code/firmware/rosco_m68k_firmware/blockdev/ata.c
@@ -318,15 +318,10 @@ static void ata_probe() {
 
 static uint8_t *berr_flag = (uint8_t*)BERR_FLAG;
 
+void TRY_DISABLE_ATA_INTERRUPT(volatile uint16_t *idereg);
+
 void ata_init() {
-    INSTALL_TEMP_BERR_HANDLER_C(&&post_write);      // Yes, this is a GCC extension.
-                                                    // No, I do not care. :D
-
-    // Disable interrupt
-    idereg[ATA_REG_WR_DEVICE_CONTROL] = 0x0002;
-
-post_write:
-    RESTORE_BERR_HANDLER();
+    TRY_DISABLE_ATA_INTERRUPT(idereg);
 
     FW_PRINT_C("Initializing hard drives... ");
 
--- a/code/firmware/rosco_m68k_firmware/blockdev/ata.c
+++ b/code/firmware/rosco_m68k_firmware/blockdev/ata.c
@@ -316,16 +316,16 @@ static void ata_probe() {
 // This means ata_init **must** be called before any video devices are
 // initialized!
 
-extern void INSTALL_TEMP_BERR_HANDLER(void);
-extern void RESTORE_BERR_HANDLER(void);
 static uint8_t *berr_flag = (uint8_t*)BERR_FLAG;
 
 void ata_init() {
-    INSTALL_TEMP_BERR_HANDLER();
+    INSTALL_TEMP_BERR_HANDLER_C(&&post_write);      // Yes, this is a GCC extension.
+                                                    // No, I do not care. :D
 
     // Disable interrupt
     idereg[ATA_REG_WR_DEVICE_CONTROL] = 0x0002;
 
+post_write:
     RESTORE_BERR_HANDLER();
 
     FW_PRINT_C("Initializing hard drives... ");
--- a/code/firmware/rosco_m68k_firmware/duart.asm
+++ b/code/firmware/rosco_m68k_firmware/duart.asm
@@ -161,9 +161,9 @@ INITDUART_ATBASE:
     ; for this on the first access, and if we get one, just bail immediately.
     move.b  #0,BERR_FLAG              ; Zero bus error flag
 
-    move.l  $8,BERR_SAVED             ; Save the original bus error handler
-    move.l  #.POST_WRITE,SDB_STATUS   ; In case we're on 68000, give a return address...
-    move.l  #BERR_HANDLER,$8          ; Install temporary bus error handler
+    move.l  $8,BERR_SAVED                   ; Save the original bus error handler
+    move.l  #.POST_WRITE,BERR_CONT_ADDR     ; In case we're on 68000, give a return address...
+    move.l  #BERR_HANDLER,$8                ; Install temporary bus error handler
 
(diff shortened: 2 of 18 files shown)


code/firmware/rosco_m68k_firmware/blockdev/ata.c
```diff
@@ -316,16 +316,16 @@ static void ata_probe() {
...
 
 void ata_init() {
-    INSTALL_TEMP_BERR_HANDLER();
+    INSTALL_TEMP_BERR_HANDLER_C(&&post_write);      // Yes, this is a GCC extension.
+                                                    // No, I do not care. :D
 
     // Disable interrupt
     idereg[ATA_REG_WR_DEVICE_CONTROL] = 0x0002;
 
+post_write:
```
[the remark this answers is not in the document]
0xTJ: Particularly, if the compiler decides to change the stack pointer here or somewhere else that `INSTALL_TEMP_BERR_HANDLER_C` would be called in the future, it wouldn't get restored. Registers could also have unexpected values (especially depending on optimization level).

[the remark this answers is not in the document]
0xTJ: Although I don't love it as a solution, one alternative would be to create a C macro like `PROBE_MEMORY_WRITE_W(addr, value)`, which uses inline asm to probe the assembly way.

This would also allow for the 68000 continuation address to be stored in a register instead of in `SDB_STATUS`. This would require a bit of extra case in assembly code that probes (so I'm not pushing strongly for this, it's just something that came to mind), but it will require fewer memory accesses while probing.

code/firmware/rosco_m68k_firmware/main1.c
```diff
@@ -230,10 +230,16 @@ if (!have_video) {
...
+#ifdef ROSCO_M68K_ATA
     ata_init();
+#endif
     INSTALL_BLOCKDEV_HANDLERS();
 #endif
 
+    // Reset the SDB_STATUS, since we (ab)use it for various things during hardware init...
+    uint32_t *sdb_status = (uint32_t*)0x404;
+    *sdb_status = 0xC001C001;
+
```
0xTJ, line 242: I could see re-using SDB_STATUS being confusing if the probing ever gets used in the future after this point. Your comment at the declaration of `INSTALL_TEMP_BERR_HANDLER_C` does address not using it beyond early init, but it would be nice for the firmware to be still able to probe later if the need arose.

0xTJ (commented):
Looking over it, I've put in a few thoughts, mostly about how the probing is handled from C.

Moving the IDE to be optional looks good. :+1:

code/firmware/rosco_m68k_firmware/main1.c
0xTJ: This point is more minor, I don't care as much about it as the `INSTALL_TEMP_BERR_HANDLER_C`, it's just a mild preference/opinion.

code/firmware/rosco_m68k_firmware/blockdev/ata.c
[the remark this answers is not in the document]
roscopeco: Fair point, this _was_ a hack. Although I believe safe _in this specific case_, definitely had potential to create hard to debug problems down the line (the kicker for me was optimisation levels as you mentioned - jumping back to the middle of a function would likely cause all sorts of weirdness that'd be super confusing 😅 ).

I've just moved this whole bit back to assembly.

code/firmware/rosco_m68k_firmware/main1.c
roscopeco: I think the rules about not probing later are probably enough, and if the need arose to probe later we should fix it at that point, but agree using `SDB_STATUS` for this was a bit of a hack, cleaned it up by giving it a dword in `.early_data` instead.

code/firmware/rosco_m68k_firmware/duart.asm
```diff
@@ -161,8 +161,9 @@ INITDUART_ATBASE:
     ; for this on the first access, and if we get one, just bail immediately.
     move.b  #0,BERR_FLAG              ; Zero bus error flag
 
-    move.l  $8,BERR_SAVED             ; Save the original bus error handler
-    move.l  #BERR_HANDLER,$8          ; Install temporary bus error handler
+    move.l  $8,BERR_SAVED                   ; Save the original bus error handler
+    move.l  #.POST_WRITE,BERR_CONT_ADDR     ; In case we're on 68000, give a return address...
+    move.l  #BERR_HANDLER,$8                ; Install temporary bus error handler
```
0xTJ, line 166: This isn't a problem with this PR, so I'll approve regardless, but would it make sense to use `INSTALL_TEMP_BERR_HANDLER` here and in VDP probing?
Right now the code is split between doing the work (clearing `BERR_FLAG`, setting up `BERR_HANDLER`) inline and calling `INSTALL_TEMP_BERR_HANDLER`, so I think that it'd make sense to standardize on one of the two approaches.

roscopeco: That's a good shout, might as well do a bit of boyscouting while I'm in this area 👍 . Done.

0xTJ (approved):
As far as I can tell, it looks good!
I've only got a 68010 running, so I can't confirm the other CPUs, but what's there makes sense. :+1:

roscopeco:
> As far as I can tell, it looks good!
> I've only got a 68010 running, so I can't confirm the other CPUs, but what's there makes sense. 👍

Thanks for the review @0xTJ 🙌

For the sake of posterity I put a screenshot in the description with proof it boots to Kermit on the bottom three 68k models 🙂

PR #388, opened 2023-11-28
Feature/kicad 7

roscopeco:
Just updated the currently-used boards to KiCad 7 format


0xTJ (approved):
Looks good, nice to update to the latest and greatest! :+1:

PR #391, opened 2023-12-02
Fix SD Card write failures by increasing timeout

harrowm:
Increase write max wait value from 1600 to 2500.  Largest wait value required through testing on a 68020 board is 2099.  Added 20% margin to that :)


roscopeco (approved):
Looks good here, really great work on this @harrowm , a lot of behind the scenes detective work!

PR #395, opened 2023-12-10
Feature/swap tx rx again

roscopeco:
This PR does TX/RX & CTS/RTS swap for both r2 and r2020 boards. There's also some tidying up in here:

* Tie unused IP pins on DUART to ground
* Fix some silkscreen issues
* Remove ROHS markings

r2030 board is not merged, but the same changes have been made in the relevant branch.

Screenshot 2023-12-10 at 17 36 33


0xTJ:
One note I do have, but haven't blocked on it here because it was already the case, is that much of that smaller text violates JLCPCB's stated capabilities for silkscreen, both on text height and line width.

roscopeco:
> One note I do have, but haven't blocked on it here because it was already the case, is that much of that smaller text violates JLCPCB's stated capabilities for silkscreen, both on text height and line width.

Good spot, it does, but they seem to be able to print it just fine, so I'm not too worried 🙂

PR #397, opened 2023-12-16
Use a Special Temporary BERR Handler for MC68000

0xTJ:
The temporary BERR handler uses the special status word/register to determine the CPU that it's running on, so that it knows how to skip processor rerun when returning from the exception. However, on the MC68000, it will use the upper 4 bits of the stacked instruction register as this format code, which could lead to a false-positive CPU detection.

This PR moves the special-case MC68000 out from the default to be its own short handler if the SDB CPU type field is zero. Only the MC68000 is moved out since it's the only one with incompatible exception stack frames.

I also added a fallback to the handler in `BERR_SAVED` for unknown stack frame formats (since this isn't MC68000 anymore) and for non-data faults on MC68020/MC68030.

I can only test on MC68010, so I haven't confirmed that these changes work on the other CPUs. The MC68000 routine *should* work since it's just the default-case code pulled out, without popping the stacked `D0`.

### DIFF
--- a/code/firmware/rosco_m68k_firmware/bootstrap.asm
+++ b/code/firmware/rosco_m68k_firmware/bootstrap.asm
@@ -350,7 +350,7 @@ INSTALL_TEMP_BERR_HANDLER::
     rts
 
 
-; Restore BERR handler, after a call to INSTALL_TEM_BERR_HANDLER.
+; Restore BERR handler, after a call to INSTALL_TEMP_BERR_HANDLER.
 RESTORE_BERR_HANDLER::
     move.l  BERR_SAVED,$8               ; Restore bus error handler
     rts
--- a/code/firmware/rosco_m68k_firmware/bootstrap.asm
+++ b/code/firmware/rosco_m68k_firmware/bootstrap.asm
@@ -277,44 +277,55 @@ INITMEMCOUNT:
     rts
 
 
-; Temporary bus error handler;
+; Temporary bus error handler for the MC68000 CPU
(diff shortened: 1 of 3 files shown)


code/firmware/rosco_m68k_firmware/bootstrap.asm
```diff
@@ -325,23 +334,28 @@ BERR_HANDLER::
...
+    move.l  #BERR_HANDLER_MC680X0,$8    ; Install other MC680x0 temporary bus error handler
+    rts
+.IS000
+    move.l  #BERR_HANDLER_MC68000,$8    ; Install MC68000 temporary bus error handler
     rts
 
 
-; Convenience to restore BERR handler from C, after a
-; call to INSTALL_TEM_BERR_HANDLER.
+; Restore BERR handler, after a call to INSTALL_TEM_BERR_HANDLER.
```
roscopeco, line 353: nitpick: typo (`TEM` vs `TEMP`)

roscopeco (approved):
Looks good! Thanks!

Tested on 68000, 010, 020 and 030 - all seem to work great 😀

code/firmware/rosco_m68k_firmware/bootstrap.asm
0xTJ: Thanks for having a look! Good catch on the typo, I've fixed that.

roscopeco (approved):
Looks good to me, good spot on this one 👍

PR #399, opened 2023-12-28
Fix incorrect number of 4k7s

roscopeco:
This BOM incorrectly listed 15 4k7s (copy/paste from 010 BOM).

This PR corrects it.


0xTJ:
The change on the 4k7 resistors looks good, but I did notice that the 10k networks RN2 and RN3 only have 1 listed instead of 2.

roscopeco:
> The change on the 4k7 resistors looks good, but I did notice that the 10k networks RN2 and RN3 only have 1 listed instead of 2.

Good spot, thanks! Should be fixed now (and I've merged them into one row too).

0xTJ:
> Good spot, thanks! Should be fixed now (and I've merged them into one row too).

Is it intentional to make RN1 go back to the 10k networks, with the merging? In the previous version I think that it was 4k7 instead, while RN2 and RN3 were 10k.

roscopeco:
> > Good spot, thanks! Should be fixed now (and I've merged them into one row too).
> 
> Is it intentional to make RN1 go back to the 10k networks, with the merging? In the previous version I think that it was 4k7 instead, while RN2 and RN3 were 10k.

Yes, that's intentional :)

0xTJ:
> > > Good spot, thanks! Should be fixed now (and I've merged them into one row too).
> > 
> > 
> > Is it intentional to make RN1 go back to the 10k networks, with the merging? In the previous version I think that it was 4k7 instead, while RN2 and RN3 were 10k.
> 
> Yes, that's intentional :)

Sounds good, wanted to double-check that! :smiley:

0xTJ (approved):
LGTM! :+1:

PR #404, opened 2024-01-07
Fix pack numbers in BOMs

roscopeco:


docs/r2_030_r2.00_bom.pdf
0xTJ, line 1: There are some empty lines after the LED4,LED2 row, which don't seem to be used.

0xTJ, line 1: The package field for the MC68030RC has MC68020RC.

0xTJ (changes requested):
Looks good, just one issue and a small note. :smiley:

docs/r2_030_r2.00_bom.pdf
roscopeco: Good spot, don't know how I missed that :D 

Thanks, fixed.

roscopeco: Also good spot, thanks again, fixed :)

0xTJ (approved):
As far as I can tell, it looks good! :+1:

PR #405, opened 2024-01-07
Misc. Minor Fixes and Cleanup in Firmware

0xTJ:
I've tried to keep the changes isolated to a single commit, so feel free to push back on any changes, they can be reverted individually pretty easily. The changes are:

- Have the early-boot dummies for string-printing EFP functions still update `A0` to point to the end of the string, as documented.
- Remove the INIT_SSP constant that's not used anywhere anymore. (The initial SSP was previously set to be below the stage2 loading location instead, so this value wasn't valid anymore.)
- Tweak to remove a magic number shift for `C_DEVICE_SIZE`. (This one isn't particularly important, actually less efficient, and shouldn't apply unless the `CHAR_DEVICE` structure is grown in the future.)
- (2 commits) Move some memory layout constants out from code and into linker scripts. (The main significant one is `RAM_TOP`, since it was using an `ifd` on REVISION1X which ignored the actual ROM size and memory layout.
- Minor change that addresses a comment about it being inefficient to use sz-strings in ANSI code printing, this replaces the single-character sequences with single-character prints.
- There were two copies of `serial.asm`, one for each stage. The version in stage1 unused except for `RECVCHAR`, which I folded into `trap14.asm` (along with the other functions), since that's where the other EFP wrappers live as `FW_<...>`.
- Fix one of the linker script changes from above only making it into the 1M version.
- Fix a `Makefile` issue that I noticed while preparing the above commit. When printing the final ROM size message, `HUGEROM` was only checked against `false`. If it was empty (use default for this board revision), it would print that the firmware was built for 1M ROMs. This was wrong in the case of `REVISION1X=true` without a `HUGEROM` set.


roscopeco (approved):
Nice, looks like some really good clean up and improvement - thanks @0xTJ !

0xTJ:
Thanks for the review! :smile:

PR #408, opened 2024-01-21
Enable instruction caches on 020 and up

roscopeco:


0xTJ (dismissed):
I can't test MC68020 myself, but from Discord it sounds like it works well (and gives a bit of a speed bump), and MC68010 still works for me. The code change looks good! :+1:

0xTJ:
Though thinking about having caches being enabled, it might make sense to add a couple lines disabling caches in `warm_boot` in `warmboot.asm`. It's used for `FW_PROG_EXIT`, so a program that enables caches then exits could cause some trouble if those caches aren't disabled again before `stage2` is loaded.

roscopeco:
> Though thinking about having caches being enabled, it might make sense to add a couple lines disabling caches in `warm_boot` in `warmboot.asm`. It's used for `FW_PROG_EXIT`, so a program that enables caches then exits could cause some trouble if those caches aren't disabled again before `stage2` is loaded.

That’s a good call, we should do that. Will add 👍

0xTJ (approved):
Updated changes look good! Tested that this doesn't break on MC68010, with MC68030 shown in Discord. Hopefully Sofowara FaFluru is a thing of the past! :+1:

PR #411, opened 2024-01-25
Fix XANSI cursor no longer visible

danodus:
Fixes the XANSI cursor no longer visible since the keyboard support has been added.


XarkLabs (approved):
Worked nicely for me testing this PR.  Good stuff. 👍

PR #412, opened 2024-01-29
Restore use of BERR_CONT_ADDR in Xosera probing for 68000

0xTJ:
In #411, the writing a continuation address to `BERR_CONT_ADDR` was removed, which should cause issues if the Xosera is not present on a system with an MC68000.

I'm not aware of a reason to have this removed, this PR adds back in that operation.


roscopeco (approved):
Looks good to me, thanks @0xTJ 👍

PR #413, opened 2024-02-05
Feature/keyboard commercial docs

roscopeco:


0xTJ (commented):
They both look quite good! The only comment I have is that the link on page 8 of the manual appears as purple (like a visited link), just in the PDF. When I open the Word file it's blue, but in the PDF it's purple. It's the only link in the document that's like this.

Does it appear like that for you too?

roscopeco:
> They both look quite good! The only comment I have is that the link on page 8 of the manual appears as purple (like a visited link), just in the PDF. When I open the Word file it's blue, but in the PDF it's purple. It's the only link in the document that's like this.
> 
> Does it appear like that for you too?

Hmm, it _does_ appear like that for me, but weirdly only in the PDF - in the original Word document it's blue like the other links 🤔 

The manuals are printed monochrome so I can't really tell if it looks different there, but I don't think so.

0xTJ (approved):
Approving since the purple link is a minor issue, but happy to hit the button again if it's changed too. :+1:

PR #414, opened 2024-02-06
Add NOP between INITMEMCOUNT write and read

0xTJ:
As seen by roberts7531 in Discord, there appears to be a situation in which the data held on the 32-bit bus of the MC68020 and later CPUs could cause the firmware memory test to have false-positives.

If no additional bus cycles occur between the write and the read (thanks to pipelining), the floating data bus could effectively cache the value, leading to memory being detected where there isn't any.

While, in theory, `DTACK` shouldn't be activated if a region is not backed by a memory device, in practice this routine is used to probe for expansion RAM, where the "legacy" expansion cards require the mainboard to generate a `DTACK` for the entire expansion RAM range.

This wasn't previously seen on the Classic '020 and '030 (because they use a 16-bit bus) or on the Pro prototypes (which didn't generate a `DTACK` for unused memory space).

This PR attempts to resolve this possible issue by adding a `NOP` between the write and the read. From the CPU manuals, this "forces instruction and bus synchronization by freezing instruction execution until all
pending bus cycles have completed." This seems to delay the fetch of the read instruction until after the write, so that the floating bus can't just hold the written value between bus cycles.

### DIFF
--- a/code/firmware/rosco_m68k_firmware/bootstrap.asm
+++ b/code/firmware/rosco_m68k_firmware/bootstrap.asm
@@ -257,7 +257,8 @@ INITMEMCOUNT:
     move.l  #.BLOCKSIZE,A0
 .LOOP
     move.l  #.TESTVALUE,(A0)
-    move.l  (A0),D0
+    nop                                 ; NOP freezes instruction execution until pending...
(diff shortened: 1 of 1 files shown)


roscopeco (approved):
Looks good, thanks @0xTJ 🙂

PR #415, opened 2024-02-20
Avoid duplication between 64k and 1MB linker scripts

0xTJ:
The main purpose of this PR is to split common firmware linker script parts out from the ROM-specific parts (and make it a bit prettier). This does make the linker scripts a bit bigger (though IMHO simpler), so I'm very open to thoughts on it.

In the ROM-specific linker script, the is a `MEMORY` statement for the board's address space, as well as one that specifies the (minimum) installed memory size. This is split apart to differentiate between the total ROM memory space (including mirrors) and the actual space available in the ROM.

In the common linker script (`INCLUDE` from the ROM-specific linker script), I've there is a `MEMORY` for the memory areas defined in the interface reference.

I've also cleaned up the `ALIGN` statements, letting sections provide the minimum required alignment. This required dropping the assumption that `.data` is at `_code_end` in ROM, since there might be alignment padding now. Instead, a `_data_load_start` symbol is provided. The vasm flag `-align` is also added to make sure that data statements will implicitly align (assembly instructions and C code/data is already aligned in the object sections).

I've copied the style of changes over the r68k, but wasn't sure if I should `INCLUDE` the common file from it, so I kept it as a single file.

This has been tested with 68010 on an r2  and an r1.2 with BIGROM (actually using a HUGEROM adapter, but the memory space is kept as BIGROM and the BIGROM image is repeated across the entire flash chips).


roscopeco (dismissed):
Looks great, thanks @0xTJ !

0xTJ:
> Looks great, thanks @0xTJ !

Thanks for the review! Before merging a did a quick check and noticed that a sign (on a 0-value, so no actual effect) was flipped. I added a commit to fix that, just in case it causes a problem in the future.

roscopeco (approved):
Looks good 👍

PR #418, opened 2024-03-18
Fix debug trap handler beyond MC68010

### LINKED ISSUE #417: Debug trap handler gets incorrect values
The debug trap handler is not fully compatible with 020 and 030 stack frames.

This causes (e.g.) fault address for bus error to be incorrect.

Check 🧐 and fix 🛠️

0xTJ:
This PR fixes #417 by adding cases to `debug_stub` for exception formats other than MC68000 and MC68010. If it is one of the exceptions that has a fault address (determined according to which handler was entered, as before) the code has a series of checks to select the current location in the exception stack frame to print as the fault address.

First it checks the SDB for CPU type, since MC68000 isn't compatible with the format of the others, the it checks the format field in the stack frame. Based on these checks, it branches to one of three instructions (MC68000/MC68010, MC68020/MC68030, and MC68040) to copy the fault address to the correct spot on the stack.

I'm open to feedback on how it's been implemented, and thoughts on if the different formats should use a table of offsets instead.

This has only been tested on MC68010.


roscopeco (approved):
LGTM, thanks @0xTJ 🥳

PR #420, opened 2024-03-21
Make LED respect SYSFLAGS and don't write to OPCR on DUART SPI read

0xTJ:
This PR includes the changes in #419 (see that PR for details on those changes) as well as additional ones to prevent the GPIOs from being incorrectly changed when the LEDs are controlled and SYSFLAGS sets them as not available to the firmware.

It adds checks to `spi_send_buffer` and `spi_read_buffer` in dua_spi_asm.asm to only set/clear the red LED if the corresponding SYSFLAGS bit is set.

There were also cases where the SYSFLAGS were and-ed with the value to set in the MFP GPDR, which seemingly would clear those bits, instead of leaving them untouched. I've changed this here to use RMW instructions with an already-masked value to directly only update the bits that are allowed by the SYSFLAGS.

I have only tested the behaviour on r2 boards with DUART, not on the r1.x with MFP. EDIT: I have now also tested that the system still works with these changes on a HUGEROM r1.23 with MFP, but haven't thoroughly tested how the LEDs behave.


XarkLabs (approved):
This looks good to me and seems better than current behavior.

PR #423, opened 2024-04-07
Fix #422

### LINKED ISSUE #422: `start_serial` incorrectly handles `EFP_PROG_EXIT` vector
Briefly:

The way the init code in `start_serial` uses the `EFP_PROG_EXIT` vector to handle program exit is flawed in two ways:

* The original handler is stored in a program variable
* The code relies on `kmain` returning to restore the original handler

This is generally "fine" but breaks when user code does "weird things" like chaining to another user program (as is the case with `sdfat_menu`). In this case, the program variable is likely to be overwritten, and the original handler is likely to never be restored, leaving it as a dangling pointer to code that no longer exists when chained code (or anything run following warm boot) calls it.

Since user programs (like `sdfat_menu`) should be able to do this, this should be fixed in the libs. The obvious (and perhaps wrong) way to do this would be to make `exit`, `atexit` handlers and the init code itself work like it does in the new(lib) startup code: https://github.com/rosco-m68k/newlib-4.4.0.20231231/blob/062620702eccd78b1f19ea7f6d4a05bdbbfb0f14/libgloss/rosco/crt0.asm#L69

For full context, see this thread: https://discord.com/channels/698525682199822367/711035186170822727/1226460649673064449

roscopeco:
This should fix #422

### DIFF
--- a/code/software/libs/src/cstdlib/stdlib.c
+++ b/code/software/libs/src/cstdlib/stdlib.c
@@ -5,18 +5,26 @@
 #define LONG_MIN -2147483648
 #define LONG_MAX 2147483647
 
-void exit(int status)
-{
-  // Jump to EFP_PROG_EXIT when quit
-  __asm__ __volatile__ (
-      "moveal 0x490.l, %a0\n\t"
-      "jmp %a0@\n\t"
-  );
-}
+// THIS IS NO LONGER DEFINED HERE, BUT IN start_serial's init.S, BECAUSE
+// the implementation is tightly coupled with `start_serial` anyway...
+//   
+// void exit(int status)
+// {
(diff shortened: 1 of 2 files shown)


0xTJ (approved):
Looks good to me! :+1: Makes sense for the C runtime stuff be "inside" the program, instead of the lower-level `EFP_PROG_EXIT` jumping back out to the C cleanup.

roscopeco:
> Looks good to me! 👍 Makes sense for the C runtime stuff be "inside" the program, instead of the lower-level `EFP_PROG_EXIT` jumping back out to the C cleanup.

Thanks for looking @0xTJ 🥳

PR #425, opened 2024-04-07
Removing GPIO lib and all dependents

### LINKED ISSUE #392: Software GPIO Directly Uses MFP
Currently, the GPIO software library is fixed to use the MFP both for Arduino-like GPIO and for SPI. This library should be fixed to be more flexible and/or call into the firmware TRAPs.

roscopeco:
This _will_ fix #392 , but kinda does it via "nuclear option".

Willing to discuss what's being removed here, or whether we should take an alternative approach....


0xTJ (dismissed):
I'm approving, because at this point I think it makes sense. Rev. 1 had more "real" GPIO (especially when they were just broken out on a larger header, before the SPI header was added), but rev. 2's pins are much more fixed-function, so I think it makes sense to act as if it has no GPIO, rather just having a baked-in SPI interface, LEDs, and UART flow control (if used).

It may also make sense to add notes to the firmware Interface Reference deprecating the system-use GPIO flags in the SDB (at least for the non-LED pins). Moving from MFP to DUART it's gotten clunky given that the bits have shifted around, and it's only currently respected everywhere for the LEDs, not for SPI. If those changes are made I'm happy to re-approve.

0xTJ:
We can always come back and resurrect these files from Git if a better GPIO system is implemented.

roscopeco:
> It may also make sense to add notes to the firmware Interface Reference deprecating the system-use GPIO flags in the SDB (at least for the non-LED pins). Moving from MFP to DUART it's gotten clunky given that the bits have shifted around, and it's only currently respected everywhere for the LEDs, not for SPI. If those changes are made I'm happy to re-approve.

Good call @0xTJ , made some doc changes :)

0xTJ (approved):
Thanks for the change, looks good! :rocket:

PR #426, opened 2024-04-07
Update designators to RUN/RESET in current BOMs

### LINKED ISSUE #409: Update BOMs to specify “RUN” / “RESET” as LED designators
Currently the BOM lists the run and reset LEDs as LEDn, which is confusing as these do not appear on the silkscreen.

BOM should be updated for consistency with boards.

roscopeco:
Fix #409


docs/r2bom.pdf
0xTJ, line 1: Is the change from `E2,5-5` to `E2,5-56` intentional for `C11,C19,C18,C17,C16,C15,C14`?

PR #427, opened 2024-05-12
Feature/smd birdseed

roscopeco:
Onboard FTDI and SMD birdseed for r2 010.

This board is tested and has gone to production. Should probably get the changes into develop here :D


0xTJ (approved):
Haven't looked every bit, but since they're working and out there it should be good. :smile:

roscopeco:
> Haven't looked every bit, but since they're working and out there it should be good. 😄

Thanks @0xTJ 👍

PR #429, opened 2024-05-20
Feature/fix d3

roscopeco:
Also removing old JLC plugin files (no longer works with KiCad 8). BOM and CPL moved to CAMOutputs


0xTJ (approved):
LGTM :+1: A nice little bit of tidying!

PR #431, opened 2024-05-20
Feature/newlib toolchain tests

roscopeco:
**WIP Removed 2024-10-04**

I think it's probably worth merging this, if only as an example (and a test in CI that we can build software with the newlib toolchain).


0xTJ (approved):
Looks good to me. :)
There are a couple declarations in libc_gdb that aren't needed, but they don't hurt things either.

roscopeco:
> Looks good to me. :) There are a couple declarations in libc_gdb that aren't needed, but they don't hurt things either.

Thanks for checking @0xTJ 🥳

PR #432, opened 2024-05-22
Link stage2 against stage1

0xTJ:
The main goal of these changes is to allow firmware stage2 to be linked against stage1, so that we can eliminate some duplicated code. I noticed in the code a comment that math functions like `divmod` were duplicated (some machine print functions were too), so figured I'd clean that up.

This PR does this by first doing partial linking on the stage1 objects, producing an ELF file where most addresses have been resolved to absolute values, but some (stage2 ZIP start and end) still need to be relocated. Then stage2 is linked, using that ELF as only a source of symbols, then compressed and turned into an object. Finally the top-level Makefile links together the stage1 and stage2 objects.
As the kernel was already been linked into a static library with its own Makefile, I've also pulled it out to the same level as stage1 and stage2.

There were a number of places with duplicated Makefile contents, so I've also moved that out to a `common.mk` file.
The part I've had the hardest time in figuring out is how to pass around config options. What I've currently settled on is `export`ing them all in the top-level Makefile, to be used by the `stage1` and `stage2`, but I'm open to thoughts on if they should be moved to a "`config.mk`" that gets included in each Makefile.

Another alternative would be to have the top-level, stage1, and stage2 all be in one `make` execution, `include`-ing instead of using recursive make.

The initial changes kept the original structure, without pulling so many things out of the `Makefiles`, so if this is too much of a change, I'm also happy to move back to that, and just keep the partial linking changes. I've got this as WIP for now to gather some thoughts.


roscopeco:
I'm not against this @0xTJ , although it's a large change on the face of it, it's relatively light touch in most places and the changes all look reasonable. The build seems to work fine in the testing I've done, and the resulting ROMs seem good too, so I'd vote for moving forward with this.

Just as a question, I noticed while building that I now need `ROM_CHIP_SIZE` as well as `ROMDEVICE` if building non-040 ROMs. No problem with that but just curious if it's new, or if I've just missed it until now?

0xTJ:
Thanks for having a look! I'll check and button things up, and move it out of being a draft.

> Just as a question, I noticed while building that I now need `ROM_CHIP_SIZE` as well as `ROMDEVICE` if building non-040 ROMs. No problem with that but just curious if it's new, or if I've just missed it until now?

What `ROMDEVICE` value are you using for the build? It should handle `AT28C256`, `SST39SF010A`, `SST39SF020A`, and `SST39SF040`, though if there's another chip that you're using we can add that or add some other default.

roscopeco:
> Thanks for having a look! I'll check and button things up, and move it out of being a draft.
> 
> > Just as a question, I noticed while building that I now need `ROM_CHIP_SIZE` as well as `ROMDEVICE` if building non-040 ROMs. No problem with that but just curious if it's new, or if I've just missed it until now?
> 
> What `ROMDEVICE` value are you using for the build? It should handle `AT28C256`, `SST39SF010A`, `SST39SF020A`, and `SST39SF040`, though if there's another chip that you're using we can add that or add some other default.

Awesome! Great stuff :) 

Ah ignore me, I was using:

```
ROMDEVICE=SST39SF020 make clean all
Makefile:88: *** === Unknown ROMDEVICE has been set without ROM_CHIP_SIZE.  Stop.
```

(So, missing the `A` off the end - adding that and it works fine :) ).

0xTJ:
> (So, missing the `A` off the end - adding that and it works fine :) ).

Ah, easy mistake to make :smiley:. I find it odd that they added the A to the ...010 and ...020 (I assume to differentiate them from similar devices from an old generation), but didn't add it to the ...040.

roscopeco (approved):
LGTM 👍

0xTJ:
Thanks for having a look!

PR #433, opened 2024-06-18
Update 2.2 docs

roscopeco:


0xTJ (approved):
LGTM! :rocket: 
I was confused by that C16, but forgot about it when I got the system up and running. :sweat_smile:

PR #436, opened 2024-07-21
Document MAME serial terminal

roscopeco:
Add a short doc on running MAME with a connected terminal program instead of the built-in


0xTJ (changes requested):
Very nice write-up! However, for some users (DoctorWkt on Discord, as well as myself), the settings don't seem to be saved when using the command line `./sbc rosco_classicv2 -window -rs232a null_modem -bitb socket.localhost:1234`.

For me, while I was able to work around this with the "Reset System" menu option, as discussed in my first message on Discord about this, a cleaner solution seems to be removing the `-bitb socket.localhost:1234` option when initially configuring the baud rate. This caused the settings to be remembered, and used later when the `-bitb socket.localhost:1234` is included again.

roscopeco:
> Very nice write-up! However, for some users (DoctorWkt on Discord, as well as myself), the settings don't seem to be saved when using the command line `./sbc rosco_classicv2 -window -rs232a null_modem -bitb socket.localhost:1234`.
> 
> For me, while I was able to work around this with the "Reset System" menu option, as discussed in my first message on Discord about this, a cleaner solution seems to be removing the `-bitb socket.localhost:1234` option when initially configuring the baud rate. This caused the settings to be remembered, and used later when the `-bitb socket.localhost:1234` is included again.

Nice, thanks @0xTJ . Should be fixed now :)

0xTJ (approved):
LGTM! Always nice to have more documentation. :)

PR #440, opened 2024-08-03
Fix an issue whereby we stomped .early_data which broke character devs

roscopeco:
### To replicate:

* Add `#include DEBUG_KEYBOARD_DETECT` at the top of `stage1/keyboard.c`
* Try to run keyboard - see `Error: Insufficient character devices`

### The issue:

`DEVICE_COUNT` is in the `.earlydata` section, which gets rolled in to `.data` and so is copied during `linit`. 
**However** it's already been populated before that, during DUART or MFP init - and copying it sets it back to zero (by copying its "source" data from ROM). This happens because the data copy just starts from code end, which is right where `.earlydata` is.

**Note** Although named `.earlydata` this section is poorly named - it's not really data, it's bss, but we also don't want it zeroed during `linit` either, since we've already set it up before `linit` is called. This is why it's called `.earlyXXX`.

### The fix

Just moved the section up after bss in KRAM, and made it not be part of `.data` in the link script.

I could also have made the copy a bit more robust and had it start from `.data_start` or whatever that symbol is called, but this way seemed to make it more clear that, despite the name, this _isn't_ data and doesn't belong in its section.

### DIFF
--- a/code/firmware/rosco_m68k_firmware/rosco_m68k_firmware_common.ld
+++ b/code/firmware/rosco_m68k_firmware/rosco_m68k_firmware_common.ld
@@ -36,7 +36,6 @@ SECTIONS
 
   .data : ALIGN(4)
   {
-    *(.early_data*)
     _data_start = .;
     _data_load_start = _data_start - ADDR(.data) + LOADADDR(.data);
     *(.data*)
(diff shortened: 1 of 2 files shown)


0xTJ (approved):
I'm confused at what's going wrong that's causing the issues. The data copy routine starts at `_data_start`, which is seemingly set to the point after `.early_data` in the linker script. But looking at the resulting map file, `_data_start` is set to the beginning of the `.data` output section, instead of the .data input section. `DEVICE_COUNT` and friends are shown as being in the `.data` input section, and being on/after `_data_start`.
Regardless, this does fix the issue, but I must be missing something about how ld works.

roscopeco:
> I'm confused at what's going wrong that's causing the issues. The data copy routine starts at `_data_start`, which is seemingly set to the point after `.early_data` in the linker script. But looking at the resulting map file, `_data_start` is set to the beginning of the `.data` output section, instead of the .data input section. `DEVICE_COUNT` and friends are shown as being in the `.data` input section, and being on/after `_data_start`. Regardless, this does fix the issue, but I must be missing something about how ld works.

Yeah, looking at the code I too am not entirely sure why this was breaking, but this does fix it and IMO it makes more sense to have that `.early_data` _not_ be in the `.data` section. 

Thanks for the review :)

Issue #18, opened 2020-03-28
Missing footprints for memory chips.


paulvdhoeven:
First: Just pulled your changes from github. 
Normally my git knowledge stops at making clones, but I have intention of learning more of git.

> paul@dualcore:~/projects/kicad/zzz_prut/rosco_m68k$ git pull
> remote: Enumerating objects: 23, done.
> remote: Counting objects: 100% (23/23), done.
> remote: Compressing objects: 100% (20/20), done.
> remote: Total 23 (delta 5), reused 13 (delta 3), pack-reused 0
> Unpacking objects: 100% (23/23), done.
> From https://github.com/roscopeco/rosco_m68k
>    4ef7063..7de5cf8  master     -> origin/master
> Updating 4ef7063..7de5cf8
> Fast-forward
>  .gitignore                               |   1 -
>  README.md                                |  69 ++++-----
>  design/r0/r0-kicad-import/rosco_m68k.pro |  43 ++++++
>  design/r1/kicad/rosco_m68k.pro           | 249 +++++++++++++++++++++++++++++++
>  images/r1-populated.png                  | Bin 0 -> 3593232 bytes
>  5 files changed, 322 insertions(+), 40 deletions(-)
>  create mode 100644 design/r0/r0-kicad-import/rosco_m68k.pro
>  create mode 100644 design/r1/kicad/rosco_m68k.pro
>  create mode 100644 images/r1-populated.png

I can now work with it as a normal KiCad project.
First thing I noticed is that footprints for the memory chips were not in the schematic.
You can easily fix this by:
**Pcbnew / File / Export / Footprint association (.cmp) File**

Then go back to Eeschema, and import that file:
**Eeschema / File / Import / Footprint Association File**

I've also used your project as an example of "vector buses" in:
https://forum.kicad.info/t/lots-of-nets-how-to-organize/21967/11

roscopeco:
Awesome, thanks for the tip 👍 I'll get that fixed asap. 

Interesting reading that page, I've read about global vs hierarchical labels. I decided to go with Global labels for now, but might re-visit that later.

PR #441, opened 2024-08-08
firmware/trap14.asm: Add RECVCHAR for FC == 17.

DoctorWkt:
This is the change to the r68k firmware that gets adventure to work.


code/tools/r68k/firmware/trap14.asm
```diff
@@ -52,6 +52,7 @@ TRAP_14_HANDLER::
     dc.l    .EPILOGUE                   ; FC == 14 ; RESERVED
     dc.l    .EPILOGUE                   ; FC == 15 ; RESERVED
     dc.l    .DEV_CTRL                   ; FC == 16 ; DEV_CTRL if so...
+    dc.l    .RECVCHAR                   ; FC == 17 ; RECVCHAR if so...
```
roscopeco: Could we change the comment to indicate that this is the `INPUTCHAR` trap, just to save confusion in the future? Also, should we add function `18` (`CHECKINPUT`) while we're here? It should just be another one-line change (it's the same as the `CHECKCHAR` call on r68k).

roscopeco (changes requested):
Thanks for the PR! 🙂 

I just left a small comment about changing a comment in the code, and adding the other related missing syscall.

DoctorWkt:
Got it, my bad. I'll close this and re-open it with a better comment and the #18 code.

Issue #17, opened 2020-03-28
Missing KiCad project file.


paulvdhoeven:
I just cloned your project because I am curious what you made of it in the mean time.

Then I discovered that the KiCad project file is missing. Without it KiCad can not be opened in project mode and the link between schematic and PCB is broken. The most logical explanation is that you forgot to put this file in git.

I also saw you entered blue text strings into the title block:
![image](https://user-images.githubusercontent.com/48927628/77823451-978e2980-70fb-11ea-8e57-f11756204d36.png)

These texts should be added via:
**"Eeschema / File / Page Settings"**

roscopeco:
Ah, many thanks (again!). I'd added `.pro` files to `.gitignore` for some reason, I've fixed now and the files are in `master`.

Also, noted your suggestion about the title block texts - I'll fix that soon.

PR #442, opened 2024-08-08
r68k firmware: added #17 INPUTCHAR and #18 CHECKINPUT

DoctorWkt:
I fixed the comments and added #18.


roscopeco (approved):
Looks good to me @DoctorWkt , thanks so much for this! Apologies it took me a while to get it reviewed.

PR #449, opened 2024-10-02
Fix crash on warm boot due to xansiterm wrapping its own input wrappers

### LINKED ISSUE #438: Crashes after warm boot
I have observed frequent crashes following warm boot with the new firmware.

This also appears after exiting e.g. sdfat menu with the / key to warm boot to kermit receive.

Suspect stack / bottom of RAM vector corruption.

0xTJ:
Currently, the warm-boot re-init of the XANSI terminal will cause it to wrap its own wrapper for the `INPUTCHAR` and `CHECKINPUT` routines.

This change fixed #438 by checking whether the current EFP routines are its own. If not, is does as before and wraps the existing values of `EFP_INPUTCHAR` and `EFP_CHECKINPUT` (now passed as arguments to `xansiterm_INIT`). If it does find its own routines, the code will instead pass in the previously-saved pointers from the XANSI data structure.


XarkLabs (approved):
I tested this in a ROM and it seems to totally fix the WARM_BOOT issue (due to XANSI "hooking itself" on 2nd init).

PR #450, opened 2024-10-04
-Wno-array-bounds for gcc14

roscopeco:
Just getting ahead of a change we'll need for GCC 14 due to new warnings

(commit message should reference `-Wno-array-bounds`, not `-mno-array-bounds` but 🤷 )


0xTJ:
Any thoughts on using `--param=min-pagesize=0` (version-dependent), as is done in `common.mk`, instead of disabling this error? This seems to be the recommended way of allowing accesses to memory in the first 0x1000, without ignoring other array-bounds errors that might come up.

As a side note, it may also be a good idea to add `-fno-delete-null-pointer-checks` to `common.mk`, to ensure that a dereference of address 0 isn't converted to a `TRAP`.

0xTJ (approved):
Looks good! Nice to get ahead of it. :+1:

PR #451, opened 2024-10-04
Feature/fix 446 (firmware splash/menu kermit upload check)

XarkLabs:
Makes starting kermit uploads faster and greatly reduces issues from kermit spuriously selecting items from firmware menu when trying to upload.
Adds a test for input on UART A and if so, quickly exits intro.  The loader also checks for input on UART A (as well as keyboard if present) and if it gets a kermit "packet start code" ('k', 0x01 or 0x11) it will exit immediately, skip trying to boot SD media and start a kermit upload.  This also allows a key pressed (UART A or keyboard) during into skip/splash to quickly exit and be passed to firmware menu (so if you manually hit 'k', menu skips to kermit upload).


code/firmware/rosco_m68k_firmware/stage2/boot_menu/main.c
```diff
@@ -97,7 +97,11 @@ int main(void) {
         mcReadchar();
     }
 
-    do_boot_menu(demo_menu_items, demo_n_menu_items);
+    int res = do_boot_menu(demo_menu_items, demo_n_menu_items);
+
+    puts("\033c");
```
roscopeco, line 102: Should these prints still be in here?

(FWIW I'm a bit surprised it actually builds, I thought pulling in printf and friends here would have blown up the size of the binary 🤷 ).

code/firmware/rosco_m68k_firmware/stage2/main2.c
```diff
@@ -140,12 +154,18 @@ noreturn void lmain() {
...
 #endif
 #ifdef WITH_BOOT_MENU
-        if (handle_boot_menu()) {
+        boot_menu_res menu_res = handle_boot_menu();
+        if (menu_res == RES_LOAD_OK) {
             goto have_kernel;
-        };
+        }
+        if (menu_res == RES_UART_LOADER)
+            goto start_uart_loader;
```
roscopeco, line 169: I think we'd need to skip this on MAME and just jump to the quick load (which I know is a pain)? Currently the build is failing for MAME firmware because this label isn't defined 😢

roscopeco: (Thinking more about it, it's probably not worth building the menu for MAME since no Xosera yet, but maybe worth keeping in for debugging purposes etc 🤔 ).

code/firmware/rosco_m68k_firmware/stage2/main2.c
```diff
@@ -177,6 +197,7 @@ noreturn void lmain() {
 #  if (defined SDFAT_LOADER) || (defined IDE_LOADER)
     FW_PRINT_C("No bootable media found\r\n");
 #  endif
+start_uart_loader:
```
roscopeco: Getting `main2.c:200:1: error: label 'start_uart_loader' defined but not used` for this when building without the menu (e.g. `ROMDEVICE=SST39SF010` or non-HUGEROM builds).

roscopeco (changes requested):
Looks great @XarkLabs , just a few small things I noticed when testing here :)

code/firmware/rosco_m68k_firmware/stage2/boot_menu/main.c
XarkLabs: Those were only for standalone build testing (hence why it builds)...I believe.  They could be removed.

code/firmware/rosco_m68k_firmware/stage2/main2.c
XarkLabs: Okay...it needs to build at least. 🙂

code/firmware/rosco_m68k_firmware/stage2/main2.c
XarkLabs: I will add more ifdefs.

code/firmware/rosco_m68k_firmware/stage2/boot_menu/main.c
roscopeco: Ah okay, of course, makes sense :) No probs with keeping them then :)

code/firmware/rosco_m68k_firmware/stage2/main2.c
XarkLabs: ifdefs added

code/firmware/rosco_m68k_firmware/stage2/main2.c
XarkLabs: added

roscopeco (approved):
Looks great, thanks @XarkLabs 🚀 

Please go ahead and merge 🙂

PR #452, opened 2024-10-04
Build smaller flash ROMs in CI

roscopeco:
Now building all the different flash sizes in CI. Since there are now differences in how they're built (360K ROMFS for 020 ROMs, no ROMFS or menu for 010) this seems worthwhile to highlight any issues introduced in the less-used sizes going forward.

All sizes are built as artifacts - proof of life here: https://github.com/rosco-m68k/rosco_m68k/actions/runs/11187767629/artifacts/2017970635


0xTJ (approved):
Looks good, makes sense to have those extra firmware files built! :+1:

roscopeco:
> Looks good, makes sense to have those extra firmware files built! 👍

Thanks for reviewing @0xTJ 🤩

PR #458, opened 2024-10-07
Fix #448

### LINKED ISSUE #448: Keyboard reset code needs to be more resilient
The keyboard init in the firmware needs to actively reset the keyboard. Currently it can get confused (and fail detection) if:

* The keyboard has been left in scan mode
* The keyboard has sent mouse packets that have not been completely processed

We should make sure we empty the buffers, send a bunch of NOPs, and then actively set ASCII mode and sensible defaults in the firmware init code.

roscopeco:
Make keyboard detection more resilient, and reset to a known state on boot.


0xTJ (approved):
The code looks good, and the keyboard still works well with it! :+1:

roscopeco:
> The code looks good, and the keyboard still works well with it! 👍

Thanks for testing! 🚀

PR #461, opened 2024-10-12
WIP: CH559 USB HID support in Firmware

roscopeco:
> **Note**: This PR is WIP and may never be progressed, let alone merged. I am raising it only to gather thoughts around whether this is something we should do, would want to do, and are prepared to support going forward.

## Initial integration CH559 to firmware

Enable with `WITH_USB_HID=true` on `make` command, currently this will auto-disable the splash screen audio to make everything fit in stage1...

**R2 boards only!**

* **Experimental**, may never make prime-time!
* Non-functional with Xosera (for reasons unknown, but suspect interplay between keyboard EFPs)
* Very lightly tested at the moment


roscopeco:
My 2c: we should maybe go with making the USB compatible with the rosco keyboard.

IMO the CH559 firmware on these devices is not of significant quality and does not support a useful subset of the USB HID features due to lack of two-way communication. Modifying it to support those things represents effort that in my opinion would be better spent elsewhere.

pros:

* less diverse device ecosystem to support
* No firmware changes needed
* Keyboard interface already working for mouse and keyboard
* * keyboard interface has reserved space for game pads etc.

cons:

* likely a bit more work
* CH559 devices would need a reflash
  * though this is true anyway due to baud rate
* potentially less flexible than a full HID implementation

roscopeco:
Closing, will not progress.

PR #462, opened 2024-10-26
MAME can't support PMOVE MMUSR, use TC instead

roscopeco:
This _should_ make no difference externally, but allows MAME to run an 030 without crashing (the Musashi MMU doesn't support move from MMUSR by the looks of it, resulting in a crash).

Reported on Discord by Eschaton.


0xTJ (approved):
Looks good to me, detection still works fine on hardware!

roscopeco:
> Looks good to me, detection still works fine on hardware!

Thanks @0xTJ 🤩

PR #466, opened 2024-11-17
Improve memcheck to report CPU and check Xoseram memory

harrowm:
Improve memcheck utility:
- report CPU and CPU speed from sdb block (and remove local asm code which didn't report 68030 correctly).  I used the same format/approach as the firmware as far as I can tell.
- fix issue with XoseRAM that the RAM block is adjacent to the ROM block and this caused the code to try to memcheck the ROM causing bus errors.
- Change order of reading memory blocks so that the report comes out in ascending order (:o)


XarkLabs (approved):
Looks like some needed improvements.

PR #471, opened 2025-01-19
Simplify address decoder logic

harrowm:
The logic for rom/ram enable is using A18/A19 and RW .. I dont think these are needed.
CPUSP isn't needed in PPDTACK as its in each element of the equation.


0xTJ:
These changes would bring a significant difference to the behaviour, where the entire RAM space is replaced with ROM while `/BOOT` is asserted, instead of just the lower quarter of it and only on reads. Currently, the firmware expects RAM to exist in the upper 3/4 of the RAM space (with A18 and A19 used to decode that), and for it to exist everywhere in the RAM space for writes (vectors are copied while still in BOOT mode).

harrowm:
Mm .. maybe I’ve misunderstood how long boot is asserted for .. isn’t it just the first 4 AS cycles ?Sent from my iPhoneOn Jan 19, 2025, at 5:48 PM, Thomas Jager ***@***.***> wrote:﻿
These changes would bring a significant difference to the behaviour, where the entire RAM space is replaced with ROM while /BOOT is asserted, instead of just the lower quarter of it and only on reads. Currently, the firmware expects RAM to exist in the upper 3/4 of the RAM space (with A18 and A19 used to decode that), and for it to exist everywhere in the RAM space for writes (vectors are copied while still in BOOT mode).

0xTJ:
> Mm .. maybe I’ve misunderstood how long boot is asserted for .. isn’t it just the first 4 AS cycles ?

Currently (with at least some of the recent hardware and the latest PLD), `/BOOT` is asserted until the first access to the middle 512 KB of any 1 MB region (A18 != A19), leaving the ROM vector table in memory until the early boot process is done and the vector table has been set up in RAM.

PR #477, opened 2025-04-24
Fix CPU type detection

0xTJ:
As raised by bolishagnasty on Discord, there were a couple issues in `cputype.asm`.

The setting of `VBR` was done with `D0` only having its lower word cleared, relying on calling code already having previously cleared the upper word. This PR corrects this by clearing the entire long-word.

Whether the test should do a read-from-`VBR` was raised, instead of a destructive write, but for the moment I consider this a feature. It ensures the system in a consistent state after a "hot" reboot (like in `updateflash`).

It was also pointed out that the the MC68060 detection (although this CPU isn't supported in the firmware anyways currently) should use an `Unimplemented Integer Instruction` exception for removed integer instructions (like `MOVEP`) instead of `Illegal Instruction`. This PR adds this vector as as one saved, replaced, and restored. This change has not been tested in practice.


0xTJ:
Creating this PR as draft, it currently doesn't build cleanly as the few extra bytes have pushed `_zip_end` past the current hard-coded start of `.romfs`. I've got a fix in progress as a separate PR to move this offset up (assuming that there is space to do so, if not I'll have to consider alternatives).

BoliShag:
If you're looking to save space I don't think SZCPU is used/referenced.

0xTJ:
For future-proofing, I want to have a bit more open space, since right now a compiler change could push us way over the edge. Sliding the ROMFS back seems feasible, and I don't think it'll break anything at this point, but I still have to investigate a bit.

BoliShag (approved):

BoliShag (approved):

BoliShag (approved):

roscopeco (approved):
🚀

PR #478, opened 2025-04-25
Move ROMFS to offset 0x30000 in flash

0xTJ:
This PR fixes the issue that appears when the `stage1`+`stage2` size increases. At least using the current toolchain, we're out of space between the end of `.zipdata` and the ROMFS, adding small amounts of code pushes us over the edge into running out of space.

A new symbol `_ROMFS_FLASH_OFFSET` is created to define this offset (set at ~`0x40000`~`0x30000` for now), and the other linker scripts are modified to use this value.

The header is also modified to use the `_ROMFS_BASE` symbol from the linker instead of hardcoding the address.


0xTJ:
I'll have to reconsider and possibly close this PR. A wrong number (320 vs. 360) on my end led to thinking that there was more free space than there is. :sweat_smile:

0xTJ:
It turns out that MicroPython was being built (at least partially) without optimizations. Adding `-Os` (though any level of optimization seems to be enough) brings that file down small enough to fit with lots of room to spare in a 320k ROMFS. The latest commit does this, and moves the start of ROMFS accordingly, still leaving an added 16k to the `stage1`+`stage2` vs. the original in `develop`.

I'm open to thoughts/comments on this change, especially if shrinking ROMFS is expected to have negative impacts.

roscopeco:
I don't see any particular issue with shrinking ROMFS a bit if we need the space for other things! Looks good to me @0xTJ 😃

0xTJ:
> I don't see any particular issue with shrinking ROMFS a bit if we need the space for other things! Looks good to me @0xTJ 😃

Thanks for the input! I've moved the PR out of being [WIP]. :smiley:

roscopeco (approved):
Looks good to me, thanks @0xTJ 👍

PR #481, opened 2025-06-01
Had to use git mv to rename MAME-Serial as MAME-serial

harrowm:
Fixes the file not working in the menu on the getting started pages


roscopeco (approved):
🎉

